PR #4, opened 2011-08-11
BSD support

Vagabond:
These patches make it possible to compile eleveldb on FreeBSD, NetBSD, OpenBSD and DragonFlyBSD. All the eunit tests pass and 'make check' for leveldb passes everywhere except for OpenBSD.

Unfortunately, because BSD make (except on FreeBSD) don't support the -C option, you need to tweak the rebar.config to call gmake instead. I don't know of a clean solution to this issue.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -1,0 +1,19 @@
+
+all: compile
+
+compile:
+	./rebar compile
+
+test: compile
+	./rebar eunit
+
+clean:
+	./rebar clean
+
+c_src:
+	cd c_src/leveldb; $(MAKE)
+
+c_src_clean:
+	cd c_src/leveldb; $(MAKE) clean
+
+.PHONY: c_src
--- a/Makefile
+++ b/Makefile
@@ -1,11 +1,25 @@
+# This is a stub Makefile that invokes GNU make which will read the GNUmakefile
+# instead of this file. This provides compatability on systems where GNU make is
+# not the system 'make' (eg. most non-linux UNIXes).
 
-all: compile
+all:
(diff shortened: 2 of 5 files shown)


ArtemGr:
> Unfortunately, because BSD make (except on FreeBSD) don't support the -C option,
> you need to tweak the rebar.config to call gmake instead. I don't know of a clean solution to this issue.

Hi!
The usual way to deal with this is to use the ${MAKE} variable. That way, the user uses "gmake" at the top level (or the port maintainer uses "gmake" from the port makefiles, which is pretty common) and it is propagated down the compile chain.

Vagabond:
Yes, except that in this case we're not guranteed to have $MAKE set, eg. if the user compiles at the toplevel via './rebar compile' instead of 'make'. Also, rebar doesn't seem to allow you access to $MAKE inside the pre-compile hook.

ArtemGr:
You already using a stub Makefile which calls gmake. Now, if that stub Makefile would use ${MAKE} (which is always available in a Makefile) then the whole thing would be Linux-compatible. FreeBSD users will just have to type 'MAKE=gmake ./rebar compile' instead of './rebar compile'. Just a thought.

Vagabond:
Rebased on top of master with gmake/make compatibility fixes.

Vagabond:
Pending the leveldb changes being merged upstream this pull request is in limbo:

http://code.google.com/p/leveldb/issues/detail?id=59

Vagabond:
This is stale and has been mostly resolved upstream now.

PR #5, opened 2011-08-13
Add snapshot support.

davisp:
Snapshots allow multiple read calls to act on the same version
of the database. This patch adds support for creating snapshots
and using them in each of the read calls.

I didn't realize till after I wrote this that @argv0 had done some work on a couple branches regarding snapshots. I poked through a bit but it looked like he was doing looking to support more than just a basic "read from snapshot". The one thing his code did remind me of is that this doesn't include support for the post_write_snapshot option. It could be added easily enough though.


matthewvon:
Is this pull request still valid / needed?  Or should I delete it?

PR #9, opened 2011-09-22
Lower default # of open files to 10; LevelDB default of 1k is just too hi

dizzyd:
Lower default # of open files to 10; LevelDB default of 1k is just too high for most use-cases


jaredmorrow:
+1

PR #10, opened 2011-11-25
We're using eleveldb in an escript

elbrujohalcon:
I know it's not the best solution, but it was enough for what we needed and, since it wasn't even working before... you may want to merge it.


dizzyd:
I'm happy to accept patches for the specs, but the code to load the .so from the escript won't work. Dynamic library loading code typically requires (at the operating system level) a file to be present in the filesystem so that it can memory-map the code into the process space.

elbrujohalcon:
I know, that's why in our last try we placed the .so file right next to escript. That is, just ../eleveldb.so instead of ../priv/eleveldb.so

dizzyd:
Sorry to leave this PR open so long. If you adjust the init/0 code to copy the file into a temporary path and load from there, I'll be happy to accept this patch. If you're no longer interested in this patch, let me know and I can close out the PR.

matthewvon:
Is this issue still valid / open?  Or should I close it?

PR #11, opened 2012-01-19
Remove embedded subtree of level in favor of traditional tarball; mainta...

dizzyd:
...ining patches with git-subtree was too painful


jaredmorrow:
+1

PR #13, opened 2012-02-06
Dss compaction tweaks

dizzyd:
Revamped build process for leveldb (to switch _back_ to our own git repo), plus commits necessary to get LevelDB compaction to not block for long periods of time

### DIFF
--- a/c_src/build_deps.sh
+++ b/c_src/build_deps.sh
@@ -32,7 +32,7 @@ case "$1" in
             (cd leveldb && git checkout $LEVELDB_VSN)
         fi
 
-        (cd leveldb && make)
+        (cd leveldb && make check)
 
         ;;
 esac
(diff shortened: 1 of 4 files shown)


massung:
Not sure if you care since it's a call into the leveldb c++ code, but a bogus repair call still returns 'ok' regardless:

```
eleveldb:repair("dataasdfasdf",[this,means,nothing]).
== ok
```

If this is ever meant to be available externally (via REST API or PBC, etc) then I think this would be confusing to users.

dizzyd:
That result should be based on check the result of the LevelDB call -- if LevelDB is returning status.ok, there's not much else to report. :)

This is a pretty low-level (and drastic) call to LevelDB; I don't anticipate it being externally invocable.

massung:
Aside from the passing an invalid directory returning 'ok', seems good. +1.

jonmeredith:
+1 merge

PR #14, opened 2012-02-07
Remove eunit pre-hook to fix toplevel riak builds

jaredmorrow:


jaredmorrow:
+1 Based on conversation with Dizzy via IM on 02/07

dizzyd:
+1, FWIW. :)

PR #17, opened 2012-03-16
Dss mar15 leveldb

dizzyd:
Update to latest LevelDB


Vagabond:
Looks good, has latest leveldb and builds and tests ok.

PR #19, opened 2012-03-26
reimplement get using leveldb::Get

vinoski:
In an email conversation Google suggested implement the eleveldb "get"
call via leveldb::Get rather than via iterators because they said
leveldb::Get was up to 15% faster than the iterator approach. Our
testing shows that implementing get via leveldb::Get results in
reduced latency.


dizzyd:
+1 to merge.

vinoski:
Merged.

PR #20, opened 2012-04-02
added explicit close function

moonpolysoft:
Needed the ability to close a database independent of NIF cleanup / beam GC.

### DIFF
--- a/c_src/eleveldb.cc
+++ b/c_src/eleveldb.cc
@@ -86,6 +86,7 @@ static ErlNifFunc nif_funcs[] =
     {"open", 2, eleveldb_open},
     {"get", 3, eleveldb_get},
     {"write", 3, eleveldb_write},
+    {"close", 1, eleveldb_close},
     {"iterator", 2, eleveldb_iterator},
     {"iterator", 3, eleveldb_iterator},
     {"iterator_move", 2, eleveldb_iterator_move},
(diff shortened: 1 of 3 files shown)


dizzyd:
Generally, this needs some bullet-proofing:
- If the handle can be deleted, all operations which use it should check for deletion before executing
- I think the handle may need to be ref-counted for proper behavior while it's also in-use for an iterator (??)

It would be extremely handy to have this functionality, though. Thanks! :)

moonpolysoft:
Added the necessary bulletproofing.  Any nif call on a closed database will badarg, and if there are outstanding iterators against a database then the call to close will badarg.

moonpolysoft:
Bump.

dizzyd:
Hi @cliffmoon, I'm pretty sure there are locking issues remaining in this code but haven't had a chance to dig into it. The API was, IIRC, safe to use across multiple processes, but with the addition of the close I think that safety is bjorked. Happy to be proven wrong. :)

dizzyd:
FWIW, I'm hoping to take a gander at this prior to our next Riak release.  Haven't forgotten...just bogged down. Lots of good LevelDB performance improvements on this way. :)

seancribbs:
@cliffmoon Looks like the explicit close function made it in during the Riak 1.2 release cycle. Closing this PR.

PR #21, opened 2012-04-04
to use local leveldb.tar.gz instead of pulling github repo

hatemogi:
It seems that pulling the leveldb repository is the only part requiring remote access while building Riak. I tried to build riak-1.1.1 from the source package on my linux machine which has no access to public internet. I couldn't build it due to building eleveldb requires remote access to leveldb github repository. So I changed a little to enable building Riak without remote access.


dizzyd:
This has been fixed in a slightly different way on master; the next major release of Riak will properly pre-download leveldb code.

PR #22, opened 2012-04-07
Modified c_src/build_deps.sh to remove bash dependency and add FreeBSD compatibility

jj1bdx:
Modification details:
- changed the shebang entry from /bin/bash to /bin/sh to remove unnecessary bash dependency
- changed the "pushd" to "cd", because there's no need to push pwd into directory stack in this script
- added Tuncer Ayaz's script of detecting "gmake" command so that systems which have gmake for GNU Make will use it instead of make (NOTE: $MAKE replaces "make" now).

### DIFF
--- a/c_src/build_deps.sh
+++ b/c_src/build_deps.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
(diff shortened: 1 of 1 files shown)


jj1bdx:
One modification needed in the URL. The correct URL for Tuncer Ayaz's source is
https://github.com/tuncer/re2/blob/master/c_src/build_deps.sh

PR #24, opened 2012-04-14
Use a /bin/sh workaround to fix building on solaris

Vagabond:
The FreeBSD fix recently merged breaks solaris, because the shebang was changed to #!/bin/sh.

@jaredmorrow


jaredmorrow:
+1 Builds pass on Solaris now

PR #25, opened 2012-04-20
Dss nonblocking nif

dizzyd:


c_src/eleveldb.cc
```diff
@@ -318,38 +354,60 @@ ERL_NIF_TERM eleveldb_get(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
...
-            else
-            {
-                return error_tuple(env, ATOM_ERROR_DB_WRITE, status);
-            }
+            // Identify the calling PID and make that available to the worker
+            ErlNifPid caller;
+            enif_self(env, &caller);
+            handle->write_caller = enif_make_pid(handle->write_env, &caller);
+
+            // Construct a reference that will be used to identify this specific
```
slfritchie: I don't know enough about the ref-based message receive optimization to answer this question so late at night, but ... it's my tiny understanding of the optimization that the ref is created by the Erlang side of the world, then sent in a message in a particular position in a tuple?

c_src/eleveldb.cc
```diff
@@ -318,38 +354,60 @@ ERL_NIF_TERM eleveldb_get(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
...
-            leveldb::Status status = handle->db->Write(opts, &batch);
-            if (status.ok())
-            {
-                return ATOM_OK;
-            }
-            else
-            {
-                return error_tuple(env, ATOM_ERROR_DB_WRITE, status);
-            }
+            // Identify the calling PID and make that available to the worker
```
jonmeredith, line 387: Would benefit from a comment to explain why the Eterm gymnastics is being used (erring on the side of caution)

c_src/eleveldb.cc
```diff
@@ -318,38 +354,60 @@ ERL_NIF_TERM eleveldb_get(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
...
+            // Construct a reference that will be used to identify this specific
+            // request back to the caller and take advantage of optimizations
+            // for long message queues
+            //
+            // PARANOID: copy the ref into the caller's env; not clear if this
+            // is strictly necessary
+            handle->write_ref = enif_make_ref(handle->write_env);
+            ERL_NIF_TERM ref = enif_make_copy(env, handle->write_ref);
+
+            // Kick the writer thread and wait for a response
```
jonmeredith, line 396: Clarify to say that it's the erlang code that will wait for the response

c_src/eleveldb.cc
```diff
@@ -361,6 +419,61 @@ ERL_NIF_TERM eleveldb_write(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
...
+            else
+            {
+                ERL_NIF_TERM error = error_tuple(handle->write_env, ATOM_ERROR_DB_WRITE, status);
+                result = enif_make_tuple2(handle->write_env, handle->write_ref, error);
+            }
+
+            // Send back the result of the write
+            ErlNifPid caller;
+            enif_get_local_pid(handle->write_env, handle->write_caller, &caller);
+            enif_send(NULL, &caller, handle->write_env, result);
```
jonmeredith, line 458: Docs say you can only call this from a non-NIF called thread on the SMP emulator.  We need to add similar protection in the NIF init code that we have in innostore https://github.com/basho/innostore/blob/master/c_src/innostore_drv.c#L188

jonmeredith:
Code looks correct - I'm going to run some benchmarks/load testing to check for any leaks/nastiness and I'm going to double-check the error paths one last time.

c_src/eleveldb.cc
dizzyd: You are completely right -- confirmed it with Ulf this morning. Pushed a change, accordingly. Good catch!

c_src/eleveldb.cc
```diff
@@ -318,38 +354,58 @@ ERL_NIF_TERM eleveldb_get(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
...
+
+            // Save the write-reference for the response back to the caller;
+            // enables us to use receive optimizations in the VM
+            handle->write_ref = enif_make_copy(handle->write_env, writeref);
+
+            // Kick the writer thread and wait for a response
+            enif_cond_signal(handle->write_cv);
+            enif_mutex_unlock(handle->write_lock);
+
+            // Return the reference
```
jonmeredith, line 400: Reference is passed in now

slfritchie:
Can this issue be closed, given all the Riak 1.3.0 work on the NIF for eleveldb?

PR #26, opened 2012-04-28
Add support for including leveldb source when get-deps is invoked

dizzyd:


jaredmorrow:
Tested with `riak_kv` master and an updated rebar.config pointing at this branch.   `make deps` correctly checked out the proper leveldb commit.

+1

Issue #31, opened 2012-06-29
compression unit test fails with latest eleveldb


jonmeredith:
The latest version of leveldb+Bbasho mods has changed when compression happens so the test now fails.  The intent of the eleveldb test is to check that the option gets passed down to leveldb correctly.

This https://github.com/basho/leveldb/pull/17 pull request adds LOG entries, modify the compression_test in eleveldb:compression_test () to check the flag is set correctly in the two LOG files.

PR #32, opened 2012-06-29
Fix the compression_test to check the passing of the compression option

macshonle:
This resolves https://github.com/basho/eleveldb/issues/31


jonmeredith:
Looks like the SHA for leveldb didn't get updated.

c_src/build_deps.sh
```diff
@@ -9,6 +9,8 @@ fi
 unset POSIX_SHELL # clear it so if we invoke other scripts, they run as ksh as well
 
 LEVELDB_VSN="557ee678a72cafec6642d18a83cabe0d418f908d" # June 25 merge w/ mv-write-throttle2
```
jonmeredith: Fix it right here...

jonmeredith:
Works for me.

+1 merge.

PR #34, opened 2012-07-30
block_size -> sst_block_size

chardan:
- block_size eleveldb parameter now has no effect on block_size setting
- new atom sst_block_size affects block_size_setting


matthewvon:
+1 ... but really should be someone that knows the code and erlang doing this ...

jonmeredith:
+1 merge

PR #37, opened 2012-08-15
raise bits per key from 10 to 16.  reduces false positive rate.

matthewvon:


chardan:
+1 (value should indeed not be const)

PR #39, opened 2012-08-27
Export iterator/3 so users can iterate over just keys

reiddraper:
I've been playing with leveldb iterators (not through `eleveldb:fold`), and would find it useful to be able to iterator through keys only. I don't know if there's a better way to deal with the 3rd argument to `iterator` only being able to be `keys_only`. If you pass in some other atom, it doesn't blow up, but just silently ignores that arg.

### DIFF
--- a/src/eleveldb.erl
+++ b/src/eleveldb.erl
@@ -38,6 +38,7 @@
          validate_options/2]).
 
 -export([iterator/2,
+         iterator/3,
          iterator_move/2,
          iterator_close/1]).
 


kellymclaughlin:
Don't see any problem with this. +1 to merge.

dreverri:
What happened to this change? This export is no longer available.

reiddraper:
@dreverri looks like some kind of commit-undo mistake in 025a3602a55845d1f4f3940158a627772d4e9bd0

reiddraper:
cc @matthewvon ^

PR #42, opened 2012-11-14
Add Erlang VM reduction count 'bumps' to all NIF calls

slfritchie:
Attempt to address the "stuck schedulers" problem by informing the R15 schedulers that there is more work going on by the NIFs than they believe.


chardan:
+1 (with note that I can't reproduce the scheduler issue to verify directly; passes tests)

PR #43, opened 2012-11-15
eleveldb write thread pool support

chardan:


c_src/eleveldb.cc
```diff
@@ -108,6 +91,410 @@ struct eleveldb_itr_handle
...
+          class P0, class P1, class P2, class P3, class P4, class P5>
+T *placement_ctor(P0 p0, P1 p1, P2 p2, P3 p3, P4 p4, P5 p5)
+{
+ return new(placement_alloc<T>()) T(p0, p1, p2, p3, p4, p5);
+}
+
+template <class T>
+void placement_dtor(T *& x)
+{
+ x->~T();
```
matthewvon, line 183: And if x is NULL?  Placement_dtor is substituting for "delete" which is NULL safe.

c_src/eleveldb.cc
```diff
@@ -108,6 +91,410 @@ struct eleveldb_itr_handle
...
+                leveldb::WriteOptions* _options)
+     : local_env(_local_env),
+       caller_ref(_caller_ref), pid_term(_pid_term), 
+       db_handle(_db_handle), 
+       batch(_batch),
+       options(_options)
+    {
+        if(0 == local_env)
+         throw;
+    }
```
matthewvon: Four pointers passed in, only one tested. Why?

c_src/eleveldb.cc
```diff
@@ -108,6 +91,410 @@ struct eleveldb_itr_handle
...
+ eleveldb_thread_pool(const size_t thread_pool_size);
+ ~eleveldb_thread_pool();
+
+ public:
+ void lock()                    { enif_mutex_lock(work_queue_lock); }
+ void unlock()                  { enif_mutex_unlock(work_queue_lock); }
+
+ void submit(work_item_t* item) 
+ { 
+    lock(), work_queue.push(item), unlock();
```
matthewvon: Personal curiousity (not bug), why commas instead of semicolons?

c_src/eleveldb.cc
chardan, line 183: Good catch, fixing.

c_src/eleveldb.cc
chardan: I sometimes use them (more often, lately) to indicate "grouping" rather than "separating" with ';'-- especially when written on one line as here. The intent is to show visually that the three operations always go together. It's easily argued that the function itself already does that and should be three lines of semicolon-separated code, but that's the reason.

jonmeredith: Doesn't sound very idiomatic. {} is for making blocks.

c_src/eleveldb.cc
```diff
@@ -108,6 +91,418 @@ struct eleveldb_itr_handle
...
+
+    h.work_queue.pop();
+    h.unlock();
+
+    // Submit the job to leveldb:
+    eleveldb_db_handle* dbh = submission->db_handle;
+
+    enif_mutex_lock(dbh->db_lock);
+    leveldb::Status status = dbh->db->Write(*(submission->options), submission->batch);
+    enif_mutex_unlock(dbh->db_lock);
```
matthewvon: We should discuss this lock with Jon.  It is redundant to what already exists within leveldb and does not guarantee the order of two submissions on the work_queue.  So either we have a work queue problem, or a redundant lock.

c_src/eleveldb.cc
```diff
@@ -173,10 +570,10 @@ ERL_NIF_TERM parse_open_option(ErlNifEnv* env, ERL_NIF_TERM item, leveldb::Optio
             // By default, we want to use a 10-bit-per-key bloom filter on a
             // per-table basis. We only disable it if explicitly asked. Alternatively,
             // one can provide a value for # of bits-per-key.
-            unsigned long bfsize = 16;
```
matthewvon, line 186: bad merge (likely my fault).  bfsize needs to be 16.

c_src/eleveldb.cc
```diff
@@ -173,10 +570,10 @@ ERL_NIF_TERM parse_open_option(ErlNifEnv* env, ERL_NIF_TERM item, leveldb::Optio
...
+            unsigned long bfsize = 10;
             if (option[1] == ATOM_TRUE || enif_get_ulong(env, option[1], &bfsize))
             {
-                opts.filter_policy = leveldb::NewBloomFilterPolicy2(bfsize);
+                opts.filter_policy = leveldb::NewBloomFilterPolicy(bfsize);
```
matthewvon: bad merge (likely my fault).  we want leveldb::NewBloomFilterPolicy2

c_src/eleveldb.cc
```diff
@@ -440,49 +839,67 @@ ERL_NIF_TERM eleveldb_get(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
...
+        eleveldb_priv_data& priv = *static_cast<eleveldb_priv_data *>(enif_priv_data(env));
 
-        // Traverse actions and build a write batch
-        leveldb::WriteBatch batch;
-        ERL_NIF_TERM result = fold(env, argv[1], write_batch_item, batch);
+        // Stop taking requests if we've been asked to shut down:
+        if(priv.thread_pool.shutdown_pending())
+         return enif_make_tuple2(env, ATOM_ERROR, caller_ref);
+
+        // Construct a write batch:
```
matthewvon: What if shutdown pending gets set to true after you pass that above test?

c_src/eleveldb.cc
```diff
@@ -440,49 +839,67 @@ ERL_NIF_TERM eleveldb_get(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
...
+
+            // Enqueue the job:
+            eleveldb_thread_pool::work_item_t* work_item = placement_ctor<eleveldb_thread_pool::work_item_t>(
+                                                            local_env,
+                                                            enif_make_copy(local_env, caller_ref),
+                                                            enif_make_pid(local_env, &local_pid),
+                                                            handle, batch, opts
+                                                           );
+
+            priv.thread_pool.submit(work_item);
```
matthewvon: Should priv.thread_pool.submit also hold the thread mutex so the shutdown flag can be properly tested before adding items to work queue?  If so, the function needs to return a bool or such for status.

c_src/eleveldb.cc
```diff
@@ -372,27 +787,11 @@ ERL_NIF_TERM eleveldb_open(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
 ERL_NIF_TERM eleveldb_close(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
```
matthewvon: What if there are items in the work_queue for this db?

c_src/eleveldb.cc
```diff
@@ -142,20 +537,22 @@ ERL_NIF_TERM parse_open_option(ErlNifEnv* env, ERL_NIF_TERM item, leveldb::Optio
         {
             unsigned long sst_block_sz(0);
             if (enif_get_ulong(env, option[1], &sst_block_sz))
-             opts.block_size = sst_block_sz; // Note: We just set the "old" block_size option.
+             opts.block_size = sst_block_sz; // Note: We just set the "old" block_size option.
```
matthewvon: Does the green block at the end of this line (and a few others) imply that your editor is adding whitespace to some lines?  Or cleaning the lines or trailing whitespace?

c_src/eleveldb.cc
chardan: I'm not the first person to do this, but I'm also not terribly attached to it. It's interesting that this is the line that causes reaction, and not the more frequent "free(x), x = 0" pattern. Here, semicolons it is.

c_src/eleveldb.cc
chardan: Sure-- it would be great if it could go away. It's been in there the whole time, but I was curious about it myself. It's easier to see things after the struggle of getting them to merely work is past.

c_src/eleveldb.cc
chardan, line 186: Yeah, or my fault... but just a merge thing.

c_src/eleveldb.cc
chardan: I did set tab expansion and did a "retab" with vim.

jonmeredith:
Had 20 mins to scan over the code and had a couple of observations

1) Why not simply the async interface and just do 

```
  Ref = make_ref(),
  case nif_func(Ref, Args) of
      ok ->
           receive 
                {Ref, Result} ->
                      Result;
       ER ->
            ER
   end.
```

You could even make an ?ASYNC_RESULT macro that wraps that up for you and still triggers the compiler optimization (though you'd need to check).

That way you can save a whole lot of extra case clauses and needing to make_pair() inside the C code to decide if it's good or bad or not.  Just make the result tuple in your async callback function.

2) non-ok response isn't handled in async_open

3) submit_job should probably be renamed async_write now - it isn't doing job submission, it triggers a put.

4) Is overloading operation () for thread pools a common C++ idiom?  Why not just define an 'execute' method, it is non-obvious to somebody reviewing the thread pool that dereferencing the pointer to the task is actually executing things.  We can ship with it, but unless there's a good reason for it (which should be commented) it seems like being overly clever to me.

5) the whitespace settings in whatever editor is being used are different to both the existing C and Erlang code which sticks out like a sore thumb when you open it in emacs or even the github viewer.

chardan:
Jon: replied via email (happy to re-repost my comments here, but I don't see an easy way to make them look readable). Thank you!

c_src/eleveldb.cc
```diff
@@ -92,22 +77,818 @@ struct eleveldb_itr_handle
...
+
+struct eleveldb_itr_handle
+{
+    leveldb::Iterator*   itr;
+    ErlNifMutex*         itr_lock; // acquire *after* db_lock if both needed
+    const leveldb::Snapshot*   snapshot;
+    eleveldb_db_handle* db_handle;
+    bool keys_only;
+};
+typedef struct eleveldb_itr_handle eleveldb_itr_handle;
```
matthewvon: Isn't this redundant?  Have not tested lately but I think the struct declaration works just like class declaration such that "eleveldb_itr_handle" is already a synonym for "struct eleveldb_itr_handle.  But I could be wrong.

c_src/eleveldb.cc
```diff
@@ -92,22 +77,818 @@ struct eleveldb_itr_handle
...
+    {
+        placement_dtor(batch);
+        placement_dtor(options);
+    }
+
+    work_result_t operator()()
+    {
+        simple_scoped_lock(db_handle->db_lock);
+
+        leveldb::Status status = db_handle->db->Write(*options, batch);
```
matthewvon, line 633: Never an error returned from a write operation?

c_src/eleveldb.cc
```diff
@@ -92,22 +77,818 @@ struct eleveldb_itr_handle
...
+    lock();
+     work_queue.push_back(item); 
+    unlock(); 
+
+    enif_cond_signal(work_queue_pending);
+
+    return true;
+ }
+
+ bool resize_thread_pool(const size_t n)
```
matthewvon, line 693: Race condition:  if two threads call this function at same time ... who wins?  Likely should take the lock here and have any wrapper for drain_thread_pool (drain_thread_pool_lockme) that allows external call directly to to drain function.

c_src/eleveldb.cc
```diff
@@ -92,22 +77,818 @@ struct eleveldb_itr_handle
...
+
+bool eleveldb_thread_pool::complete_jobs_for(eleveldb_db_handle* dbh)
+{
+ if(0 == dbh)
+  return true;  // nothing to do, but not a failure
+
+ db_matches m(dbh);
+
+ // We don't want more than one close operation to reshuffle:
+ ErlNifMutex *complete_jobs_mutex = enif_mutex_create((char*)"complete_jobs_lock");
```
matthewvon: I do not see the documentation claiming that enid_mutex_create will fail if two threads create mutex with same name.  Is that what this code is depending upon?

c_src/eleveldb.cc
```diff
@@ -92,22 +77,818 @@ struct eleveldb_itr_handle
...
+
+ // Stop new jobs coming in during shuffle; after that, appending jobs is fine: 
+ {
+ simple_scoped_lock l(work_queue_lock);
+ std::stable_partition(work_queue.begin(), work_queue.end(), m);
+ }
+
+ // Signal all threads and drain our work first:
+ enif_cond_broadcast(work_queue_pending);
+ while(not work_queue.empty() and m(work_queue.front()))
```
matthewvon, line 846: Two questions:
1.  you are operating on an stl::deque without holding a lock on the queue.  Are the operations guaranteed to be thread safe?  I thought the answer was always no.
2.  Next line does nothing ... what is deleting items from deque?

c_src/eleveldb.cc
```diff
@@ -92,22 +77,818 @@ struct eleveldb_itr_handle
...
+ {
+    bool state;
+
+    release_thread()
+     : state(true)
+    {}
+
+    void operator()(ErlNifTid*& tid)
+    {
+        if(0 != enif_thread_join(*tid, 0))
```
matthewvon: You should broadcast before join on every iteration.  There is no guarantee the thread in proper place to see shutdown bool or signal.  There might be a guarantee if you moved the l(threads_lock) object to precede setting both shutdown and sending the broadcast.  I would do both, this code does not need to be high performance.  Safety first.

c_src/eleveldb.cc
chardan: Yep, been in there for a bit-- removed.

c_src/eleveldb.cc
chardan, line 633: The first element here should reflect the error condition back to the user. I've made sure a more reasonable atom is also returned (in reality, the way that I'm returning is a little bit braindead and we sometimes return an extra atom-- I won't change it now (it touches quite a lot of code and shouldn't be causing a /problem/ right now) but on a future pass we might want to return "real" tuples rather than the hardcoded approach.)

```
   return std::make_pair(status.ok(), ATOM_OK);
```

...now the atom will also match the status.

PR #44, opened 2012-12-19
Mv thread direct

matthewvon:
Initial implementation of thread to thread work handoff.  Bypasses backlog queue for better performance in query operations.


chardan:
+1 (with the understanding that after we kick the tires we'll address any resource management, thread pool resize, and cosmetics later)

PR #45, opened 2012-12-20
Jfw return value rodeo

chardan:
Fix regression with riak_kv.


slfritchie:
The basics look good.  TravisCI's unit tests are passing.

One thing before final review: a second combing of the -specs in src/eleveldb.erl are in order.
1. Check that the return values from the async work items match.  Jesse and I have already gone through some but not all.
2. Add the type for the "ER" returned when the async job submission fails.  Perhaps add a `-type async_failure() :: whatever_the_ER_thingie_looks_like()` and then add `async_failure()` as an option to the return value spec for each of the async functions?

PR #49, opened 2013-01-25
Mv unordered close

matthewvon:
This is to address Solaris atomics syntax and potential GC of db before iterator referencing db.


chardan:
+1 from me, but let's wait to hear back from Joseph as well.

PR #50, opened 2013-01-27
Address race condition between queue asking for help and a single worker...

matthewvon:
... thread being able to respond to such a request.


jonmeredith:
Resolves my issue, safe code change as will recheck state of queue.

+1 merge.

PR #51, opened 2013-01-29
Let Invalid() prefetches participate in HandoffAtomic logic.  Helps AAE ...

matthewvon:
...case where SEEKs are shifting.


jtuple:
Code looks good and with these changes, the AAE eunit tests once again pass without issue. +1 merge

PR #53, opened 2013-03-04
add BinaryValue class to avoid std::string copy on Get()

matthewvon:
Add a BinaryValue class derived from leveldb::Value to enable leveldb Get()
calls to copy directly into a NIF binary and thereby avoid the old approach
of copying first into a std::string and then copying again to the binary.


matthewvon:
This was manually merged to main.  Going to delete.

PR #54, opened 2013-03-16
OpenBSD needs inc/dec_and_fetch templates for size_t

jmatthew:
This is needed to produce an eleveldb.so that loads correctly on OpenBSD.


matthewvon:
I am "finally" getting back to eleveldb work.  Is this fix still considered valid?  If so, I will merge it quickly.

jmatthew:
It's still necessary with eleveldb as shipped in riak 1.4.1.

jmatthew:
thanks!

PR #56, opened 2013-04-02
fix build on s390

lemenkov:
This fixes building on s390x. It still fails to pass the test but this might be a different issue.

http://s390.koji.fedoraproject.org/koji/getfile?taskID=1012323&name=build.log


matthewvon:
The same line was changed for BSD.  I am manually copying your change to the file and committing.  Not using this pull request, but am using the change.

PR #57, opened 2013-04-12
add get-deps and rm-deps makefile targets, clean doesn't delete leveldb

nicad:
add get-deps and rm-deps makefile targets to download/delete dependencies (leveldb), target 'clean' doesn't delete leveldb if there

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,12 @@
 
 all: compile
 
+get-deps:
+	./c_src/build_deps.sh get-deps
+
+rm-deps:
+	./c_src/build_deps.sh rm-deps
+
 compile:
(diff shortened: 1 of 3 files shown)


matthewvon:
I like this change.  I am checking with those that may have a different view of Riak build process before accepting.

matthewvon:
Scott reviewed.  Says is fine.  I will work the manual merge.

matthewvon:
The Makefile and build_deps.sh changes needed to be manually merged to "master".  Done.  Closing this PR.  The race condition was fixed long ago and now the build changes are also incorporated into master.  Thanks again to nicad.

PR #58, opened 2013-04-12
Fix race condition with thread direct work

nicad:
I can see in production a case where threads get stuck doing eleveldb operations, they just never get a response from their async get or put.

You find a test program below to reproduce the problem (does it on a production server but not on my dev laptop).
Production:
- eleveldb: master from March 5th 2013
- OS: FreeBSD 9.1-PRERELEASE
- HW: amd64 32 CPUs, 
- Erlang: r16 running with -zdbbl 16384 -swt medium -sbt tnnps -sbwt none -A 128

I tracked down the problem to a wrong assumption (or oversight) that if a thread is blocked in pthread_cond_wait and one thread does broadcast/unlock the next thread to get the lock is the one blocked, that's not a correct assumption.

The problem can happen with 3 threads:

```
thread A:
    FindWaitingThread
        sees worker thread B available, mark it as busy
        doesn't lock mutex yet

thread B takes its lock
    make itself as available
    wait on cond var

thread C 
    FindWaitingThread
        marks thread B as busy    
        takes mutex
        set work
        flag cond var

thread A can now take the lock
    overwrite DirectWork
    flag cond var

thread worker X gets out of cond_wait
    marks itself not available
    get the work of thread A, work of thread C will never be processed/replied to
```

Here is a test program that could reproduce the problem and confirmed the fix:

``` erlang
-define(READ_OPTIONS, [{verify_checksums, false}, {fill_cache, true}]).
-define(WRITE_OPTIONS, [{sync, false}]).

-record( wstate, {
        id,
        controller_pid,
        max_loops,
        loops = 0,
        ref
    }).

stuck_worker(#wstate{loops = N, max_loops = N, controller_pid = CPID} = S) ->
    CPID ! {done, S};
stuck_worker(#wstate{id = Id, loops = L, ref = Ref} = S) ->
    Key = to_binary( pid_to_list(self()) ),
    Value = to_binary(L),
    WRatio = 100,
    case L > 0 of
        true ->
            RV = to_binary( (L-1) - ( (L-1) rem WRatio) ),
            {ok, RV} = eleveldb:get(Ref, Key, ?READ_OPTIONS);
        _ ->
            ok
    end,
    case (L rem WRatio) == 0 of
        true ->
            ok = eleveldb:put(Ref, Key, Value, ?WRITE_OPTIONS);
        _ ->
            ok
    end,
    case (L rem 100000) == 0 of
        true ->
            io:format("~p ~p done ~p iterations~n", [Id, self(), L]);
        _ -> ok
    end,
    stuck_worker(S#wstate{loops = L + 1}).

stuck_test() ->
    stuck_test(10, 10).

stuck_test(NProcs, Rounds) ->
    os:cmd("rm -rf /tmp/eleveldb_stuck_test"),
    {ok, Ref} = eleveldb:open("/tmp/eleveldb_stuck_test", [{create_if_missing, true}]),
    ok = eleveldb:put(Ref, <<"abc">>, <<"123">>, []),
    {ok, <<"123">>} = eleveldb:get(Ref, <<"abc">>, []),
    not_found = eleveldb:get(Ref, <<"def">>, []),

    WorkerState = #wstate{
        controller_pid = self(),
        max_loops = Rounds,
        loops = 0,
        ref = Ref
    },
    io:format("Spawn all: ~p~n", [NProcs]),
    Workers = [ {N,  spawn_link( ?MODULE, stuck_worker, [WorkerState#wstate{id = N}] )}
        || N <- lists:seq(1, NProcs)
    ],
    io:format("All workers started, wait~n", []),
    [ 
        begin
            io:format("wait for ~p pid=~p~n", [N, Pid]),
            receive
                {done, #wstate{id = N} = Info} ->
                    io:format("worker done: ~p~n", [Info])
                %%Other ->
                    %%io:format("invalid response from worker: ~p~n", [Other])
            end
        end
        || {N, Pid} <- Workers
    ],
    ok.

%% I could create the problem with just 100 processes and 5000 updates per thread
stuck_test(100, 1000*5).
```

While I completed agree with the need to avoid blocking the Erlang VM the current implementation is too complex, I'm pretty sure there are other bugs with this mix of locks and atomic variables:
- why not make eleveldb a driver ? it pretty much works like one, it would remove one layer to optimize from erlang app/erlang VM/eleveldb custom thread pool/leveldb threads/OS. 
- should not have m_Available and m_DirectWork, just doing CAS on m_DirectWork is logically enough (setting it to NULL to indicate that the thread is available), I didn't get it to work due to my unfamiliarity with the ref counting scheme used (and didn't invest enough time)
- are the compare and swap/atomic variables really worth it ? or is it just that less code is executed than with the dequeue ? i.e. if an application gets into contention issues is it really benefiting from leveldb in the first place or should it just use an in-memory database ?

Hope I don't sound like complaining, I appreciate being able to use eleveldb, thanks for making it available. I just want it to be as good as possible.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,12 @@
 
 all: compile
 
+get-deps:
+	./c_src/build_deps.sh get-deps
+
+rm-deps:
+	./c_src/build_deps.sh rm-deps
+
 compile:
 	./rebar compile
 
--- a/c_src/build_deps.sh
+++ b/c_src/build_deps.sh
@@ -31,8 +31,15 @@ MAKE=${MAKE:-make}
 # Changed "make" to $MAKE
 
 case "$1" in
-    clean)
+    rm-deps)
         rm -rf leveldb system snappy-$SNAPPY_VSN
+        ;;
+
+    clean)
+        rm -rf system snappy-$SNAPPY_VSN
+        if [ -d leveldb ]; then
+            gmake -C leveldb clean
+        fi
         ;;
 
     test)
--- a/c_src/eleveldb.cc
+++ b/c_src/eleveldb.cc
@@ -254,8 +254,13 @@ private:
                  //  of signal to cover one other race condition
                  //  that should never happen with single thread waiting.
                  pthread_mutex_lock(&threads[index]->m_Mutex);
-                 threads[index]->m_DirectWork=work;
-                 pthread_cond_broadcast(&threads[index]->m_Condition);
+                 if (threads[index]->m_DirectWork == NULL) {
+                     threads[index]->m_DirectWork=work;
+                     pthread_cond_broadcast(&threads[index]->m_Condition);
+                 }
+                 else {
+                     ret_flag = false;
+                 }
                  pthread_mutex_unlock(&threads[index]->m_Mutex);
              }   // if
          }   // if
@@ -552,19 +557,34 @@ void *eleveldb_write_thread_worker(void *args)
             submission->RefDec();
 
             submission=NULL;
+
+            // protect with a local to avoid having compiler re-order the 2 assignment and resulting into condition check issue elsewhere
+            pthread_mutex_lock(&tdata.m_Mutex);
+            if (tdata.m_DirectWork == NULL) {
+                tdata.m_Available = 1;
+            }
+            pthread_mutex_unlock(&tdata.m_Mutex);
         }   // if
         else
         {
             pthread_mutex_lock(&tdata.m_Mutex);
-            tdata.m_DirectWork=NULL;
-            tdata.m_Available=1;
-
-            // only wait if we are really sure no work pending
-            if (0==h.work_queue_atomic)
-                pthread_cond_wait(&tdata.m_Condition, &tdata.m_Mutex);
-
-            tdata.m_Available=0;    // safety
             submission=(eleveldb::WorkTask *)tdata.m_DirectWork;
+            if (NULL==submission)
+            {
+                tdata.m_Available = 1;
+                // only wait if we are really sure no work pending
+                if (0==h.work_queue_atomic)
+                    pthread_cond_wait(&tdata.m_Condition, &tdata.m_Mutex);
+            }
+
+            if (submission == NULL)
+            {
+                submission=(eleveldb::WorkTask *)tdata.m_DirectWork;
+            }
+            else {
+                tdata.m_Available = 0;
+            }
+            tdata.m_DirectWork=NULL;
             pthread_mutex_unlock(&tdata.m_Mutex);
         }   // else
     }   // while


slfritchie:
Using 6-core and 8-core boxes (desktop-style) with Winchester disk for /tmp, I couldn't hit the race ... even when doing something like 3x or 5x as many workers procs & iterations.

However, on r1s11.bos1 in the Basho Boston colo facility, I see the race after just a few minutes:

```
% export PATH=${PATH}:/usr/local/basho/erlang/bin
% git log | head -5
commit 217791f75e77ec55bd3af29430593e3e2ff7ec56
Merge: 5f21310 abc584d
Author: Christopher Meiklejohn <christopher.meiklejohn@gmail.com>
Date:   Sun Mar 31 16:34:40 2013 -0700

% erl -pz ebin -zdbbl 16384 -swt medium -sbt tnnps -sbwt none -A 128
> [ ok = stuck:stuck_test(100, 1000*5) || _ <- lists:seq(1,100)].
```

The tell-tale output says something like this, followed by silence:

```
wait for 35 pid=<0.1187.0> at {2,34,21}
worker done: {wstate,35,<0.32.0>,5000,5000,<<>>}
wait for 36 pid=<0.1188.0> at {2,34,21}
worker done: {wstate,36,<0.32.0>,5000,5000,<<>>}
wait for 37 pid=<0.1189.0> at {2,34,21}
```

matthewvon:
The initial deque coding caused a 58% drop in throughput.  Hence the development of this ugly thing.

One of your comments mentions compiler reordering.  Would you mind expanding on your findings?  I have not yet walked the code, but am curious/concerned how Thread A and Thread C were able to both mark the same worker thread busy.

Thank you for taking time to research this and rebar issue.

nicad:
this bug is NOT due to compiler reordering, the problem is that producer threads do CAS from 1 to 0 before taking the mutex, the consumer does a direct assignment to 1 when holding the lock before releasing it in the cond_var making itself look available for other consumer threads to try to give it direct work.
Compiler re-ordering is one worry with this mix of variable assignment to avoid the shared queue but I'm more worried about just protocol errors like above. 

Do you have any information about the source of the 58% drop ? i.e. how much was contention and how much was longer code-path managing the dequeue ? just curious ...

I'm wondering if the Erlang VM could do anything better with a driver + async thread pool.

matthewvon:
First my apologies for the delayed reply.  I ran the Boston Marathon on Monday and just did not feel up to any real work yesterday.

Let me cover the easy topics first.  The choice of NIF over driver is a legacy issue.  We did discuss a rewrite for 1.3 but were too far into the schedule (i.e. already late).  Your research here may bring that topic up again for our 2.0 release.

58% performance drop seemed to be a combination of issues.  The obvious issue is that each operation that used to execute directly now goes through at least two thread changes:  once from eleveldb to leveldb and again from leveldb to eleveldb (and each thread change had to copy data again).   Linux "perf top" command/tool showed heavy contention on the the "pthread_mutex_wait" function.  This was the second issue.  That lead to the realization that having all the Erlang threads and the worker threads manipulating the same mutex was suboptimal.  Six to eight developers contributed code / algorithms for performance testing over 3 very long months.  Various optimizations were attempted both in Erlang code / messaging and in the C++ code.  The atomics in this routine and similarly in the iterator section of code are what brought performance back to a shipping point.

About this issue, again thank you for taking time both to identify it and to create a program that forces it to reoccur.  Both are invaluable.  My coworker Scott has verified the bug via your test program.

That said, I do not believe your fix addresses the root cause of the problem.  I would appreciate your eyeballs and opinions concerning lines 560 and 563 in the original code:

tdata.m_Available=1;

// ... a comment ...
if (0==h.work_queue_atomic)

The fact that you found two erlang threads believing they had each claimed ownership of the one worker thread troubles me.  I believe these two lines let that happen.  The sequence:  Worker thread B performs t_data.m_Available=1 at line 560, Erlang thread A does a CAS of 1 to 0 (taking ownership) and is suspended, Worker thread B discovers queue is NOT empty in line 563 and bypasses the pthread_cond_wait() to perform a work loop, Erlang thread A is still ready but not running, Worker thread B gets back to line 560 to set Available=1 again, Erlang thread C now has a chance to also claim this worker thread too.  One of the two m_DirectWork objects gets lost.

My work plan:
- duplicate bug on my assigned hardware using your code
- see if moving tdata.m_Available=1 inside line 563's "if" condition fixes, otherwise look again for root cause
- once root cause proven, merge in defensive code provided by your patch

Your thoughts / concerns / insights?

matthewvon:
P.S.  Did you first experience this bug with Riak or with some other application using eleveldb directly?

nicad:
Thanks for the background regarding the current implementation, it helps to think about what else might be worth considering.

I came onto this problem while using eleveldb directly, not through Riak.

I've ran my fix for two 24h periods at moderate load without any issue but it doesn't mean much, it's very likely that my fix still has bugs.

You are right, your scenario is another possible race condition in addition to the one I described (probably more likely).
So basically we can ignore the value of available, it's really an optimization to avoid taking the mutex, it serves no logical purpose given that it's reset without a CAS. 

Looking back at my patch, it's pretty ugly, I wished I had spent more time trying to get the templated CAS to work directly with m_DirectWork. 
But I think it does address the scenario you describe because both the producer and consumer code checks the value of m_DirectWork while holding the lock so it should never get magically reset.

My recommendation is to not bother with moving m_Available around and instead just get rid of it and do CAS on m_DirectWork to know if there is work or not.

In pseudo code:

producer:

```
CAS work pointer with old=NULL, new=MyTask
if failed, try another thread
if successful, lock/notify/unlock
```

consumer loop:

```
# check direct work first
submission = work;
if (submission != NULL) 
    perform_ask(submission)
    CAS work with before=submission, new=NULL (should always succeed)
else
    # no direct work, go to shared queue
    if queue.empty
        lock, cond_wait unlock (i.e. loop around)
    else
        lock queue
        submission = pop()
        unlock queue
        if (submission != NULL)
            perform_task(submission)
            submission = NULL;
            next
        endif
endif
```

I also see the comment about compiler reordering you were asking about earlier: 
"protect with a local (typo: s/local/lock/) to avoid having compiler re-order the 2 assignment and resulting into condition check issue elsewhere", 
This may be paranoid of me but I don't want to assume that the compiler won't reorder 2 distincts assignments that don't refer to each others even if declared as volatile and CAS probably involves a memory barrier.

say: 

```
volatile A, B
```

One function does:

```
if (A == something_else)
B = v
```

Another function does:

```
B = something
A = something_else
```

which may become:

```
A = something_else
B = something
```

screwing up the first function.

Locks are worrisome, CAS are scary but code with CAS+non-CAS-read/write+lock is probably doomed.

matthewvon:
I have created a new branch, mv-nicad-issue-58.  I would appreciate if you would use this branch in your environment.  It intentionally does not YET include your defensive programming.  I am looking to see if the basic flaw that enables multiple erlang threads to claim same worker is corrected.  Your hardware is superior to my best server (24 core).

The change to c_src/eleveldb.cc looks like this (and is only change in the branch):

diff --git a/c_src/eleveldb.cc b/c_src/eleveldb.cc
index b36a49a..bc58eb9 100644
--- a/c_src/eleveldb.cc
+++ b/c_src/eleveldb.cc
@@ -557,12 +557,13 @@ void *eleveldb_write_thread_worker(void *args)
         {
             pthread_mutex_lock(&tdata.m_Mutex);
             tdata.m_DirectWork=NULL;
- ```
         tdata.m_Available=1;
  
       // only wait if we are really sure no work pending
       if (0==h.work_queue_atomic)
  ```
-           {
-               tdata.m_Available=1;
               pthread_cond_wait(&tdata.m_Condition, &tdata.m_Mutex);
  -
-           }    // if
           tdata.m_Available=0;    // safety
           submission=(eleveldb::WorkTask *)tdata.m_DirectWork;
           pthread_mutex_unlock(&tdata.m_Mutex);

matthewvon:
nicad's test script looped successfully for 21 hours on simple mv-nicad-issue-58 branch.  Previously failed / locked up in a few seconds on same machine.  Would appreciate outside confirmation.

slfritchie:
Matthew, I think I've convinced myself that your change avoids this ticket's problem.  I agree with @nicad that this feels kludgy and that it'd be less kludge if m_Available were eliminated ... but using a CAS on m_DirectWork introduces a bigger change that would require more review than my atrophied C + concurrency muscles can deal with.

Using an unpatched eleveldb.cc, I've been using a 16 core box that would also hang after less than a minute of run time using nicad's test program.  On that same box and same workload (as well as with an 5x increase of worker procs and 5x more iterations per proc) and also Matthew's patch, I've been running for nearly two days without a lockup.

It looks like the bug has been found.  Matthew, were you still planning on adding defensive code similar to nicad's?

matthewvon:
m_DirectWork is a pointer.  It is valid for an Erlang thread to set this to NULL while claiming a worker thread.  See the routine submit() around line 272 of eleveldb.cc.  submit() first attempts to pass the work item directly to a worker thread via FindWaitingThread(item).  If that call fails to find an available thread, submit() puts the work item on the backlog queue.  Then submit() calls FindWaitingThread(NULL) just to remind any working thread that may have gone idle between the first call and now that it should wake up on work the queue.

Worker threads should:
1. process m_DirectWork if available
2. process backlog queue until empty
3. go back into waiting mode, setting m_Available to true.

Yes, we could play games with m_DirectWork to have represent 3 different states:
A.  thread waiting for work
B.  thread claimed and here is the direct work
C.  thread claimed and there is no direct work, skip direct to backlog queue work

It is my design choice to not overload m_DesignWork as a tristate object.  That is why m_Available exists.  Some of this last github comment should go into the code comments.  Doing that now.

nicad:
Sorry for the late reply.

Thanks for the branch, I'll try it out later today.

One way to get rid of m_DirectWork without playing games and staying true to the current design would be to enqueue a CheckQueue task instead of doing FindWaitingThread(NULL); The CheckQueueTask would just be a no-op (can even be a singleton to avoid memory alloc). 

Forgot to ask earlier, the 58% performance drop, this must have been for a CPU bound load right ? Is there anything to read about what was tried during these 3 months ? so that I don't make you repeat.

The worker thread pool seems to be shared by all instance of leveldb (if my understanding is correct, I'm new to Erlang NIFs and might be misreading the code), this is not ideal, we are using 1 leveldb instance per independent disk to drive as much I/O to it as possible, ideally each instance/drive should have its own pool so they don't affect each others.

Thinking of what a future version could look like: map 1 eleveldb thread to the erlang thread doing the open, opens to the same db name would map to the same leveldb object. I see many benefits:
- no erlang VM lock-up, the async implementation is still used, lot of code reuse 
- no need to choose a size for the eleveldb thread pool, just do it in erlang
- no extra mutex/locking done in the NIF, leveldb already does proper locking within the same process I believe
- minimize sharing between processes/threads/disks
- if leveldb becomes slow then backlog happens at the erlang level
- no need for mutex/atomic ops for ref-counted objects or metric counters (micro-optimization most likely but more in the Erlang concurrency spirit)
  I believe it means to use a resource object in NIF-speak.
  Any drawbacks to this approach ? (besides ::open() becoming more heavy-weight which is easily avoided by keeping the returned handle in the erlang process' state).

matthewvon:
- 58% drop is count of operations per second through Riak
- other coding attempted (see branches): gsb-async-nif, sbv-db-value-holder (used), jdm-batch-and-prefetch-fold, jdm-batch-and-prefetch-fold-newpool, jdb-iterate4, jdb-iterate5, sbv-alloc-iterator6, jdb-mv-iterate6, jdb-mv-iterate7.  These branches contain only code that past engineer's local performance tests and was presented to the team.  Many "micro-bench" tests looked good only to yield little to no improvement in macro benchmarks.  We only cared about macro.
- this is optimized for Riak.  Other uses may optimize differently.  Even the upcoming Riak 2.0 is expected to require adjustments at this layer again.  One pitch so far is to isolate thread handoff by leveldb database (this may be part of what you mention in your "open()" discussion.
- FindWaitingThread(NULL) is fastest way to address race condition of:
  - FindWaitingThread(work_object) is unable to find a working thread available
  - one or more working threads become available, see nothing in queue, go to sleep
  - work_object is added to backlog queue
  - FindWaitingThread(NULL) will find a now available working thread and spur it to find work_object now on queue

Without the FindWaitingThread(NULL), work_object has to wait until some other active thread finishes an in-progress task and then searches the queue.  If the number of working threads were small, there is also the possibility of work_object sitting on the queue until some other Erlang work is requested (which could be a long, long time).

nicad:
Good news: I was able to run my test program in production successfully with the patch from matthewvon.

I understand why FindWaitingThread needs to be called twice, my suggestion was to avoid making m_DirectWork a tristate object.
Just noticed, my comment had a typo 's/get rid of m_DirectWork/get rid of m_Available/g' 

My proposal goes beyond 1 thread per leveldb instance, that would not help in case of read concurrency (I'm assuming that reads take very short locks inside leveldb compared to writes) 
My proposal would be to have 1 OS thread per Erlang process and per leveldb db name. I think it will provide better concurrency and simpler NIF code (no queueing needed). Please let me know if there is a public discussion I can follow on Riak 2.0 eleveldb changes, I'm considering other enhancements which probably don't belong to this thread (will go through the branches to see what was tried, thanks for pointing them out).

Also I just realized that the eleveldb pool is shared by all erlang schedulers and by all leveldb instances, a big choke point and probably an amplified head of line effect when one leveldb instance gets slower.

Btw some reference on volatile and re-ordering (slides 25, 26): http://www.cs.uiuc.edu/class/fa07/cs498mjg/notes/par3.pdf

matthewvon:
Closing this pull request because issue was fixed slightly differently on another branch.

PR #59, opened 2013-04-21
Issue 58:  Address condition where multiple erlang threads could take ownership of ...

matthewvon:
...same worker thread ... somebody is going to lose a message.


slfritchie:
This PR addresses https://github.com/basho/eleveldb/pull/58

PR #63, opened 2013-06-19
add support for new leveldb Option variable verify_compactions

matthewvon:


engelsanchez:
Consider adding the new option to the type specs in eleveldb.erl for max OCD (mimic the other option)

engelsanchez:
Other than that, :+1:

PR #66, opened 2013-07-26
quick copy/paste to move threading stuff away from eleveldb.cc API stuff.

matthewvon:
No Code Changed!  Simple copy paste of threading stuff to its own source file ... then cleanup of other misc. to get things to compile and unit test again.


engelsanchez:
I don't see deletes here and was expecting some if code was simply moved. Or am I missing something?

c_src/eleveldb.h
```diff
@@ -50,5 +50,10 @@
 
 } // namespace eleveldb
 
+namespace {
```
engelsanchez: Now that this is in a header file, having it in an anonymous namespace looks kind of dirty. Consider adding it to the leveldb namespace. Also, wouldn't it make sense to have this in threading.h?

engelsanchez:
Beyond that comment on the anonymous namespace and location of the max # of threads, I have nothing. Looks good, compiles and passes tests. :+1: :dancer:

engelsanchez:
Looks good to me. I re-checked compilation and tests :+1:

PR #67, opened 2013-08-08
Fix spec for async_get

rzezeski:
This is an async get, the previously defined spec was conflating the
message being sent to the request pid with the term actually returned
by the function call.  The later can only be the atom `ok` or a
`badarg` exception.


rzezeski:
This is replacement for #47.

matthewvon:
he tells me it works.

PR #68, opened 2013-08-09
Mv clean overlaps

matthewvon:
Delving into the cachetest.erl unit test revealed two things:  the test is a royal pain, and due to its pain we missed an actual cache memory leak.  This branch works the royal pain and the leveldb mv-clean-overlaps works the memory leak.


c_src/build_deps.sh
```diff
@@ -8,7 +8,7 @@ if [ `uname -s` = 'SunOS' -a "${POSIX_SHELL}" != "true" ]; then
 fi
 unset POSIX_SHELL # clear it so if we invoke other scripts, they run as ksh as well
 
-LEVELDB_VSN="1.4.1"
+LEVELDB_VSN="mv-clean-overlaps"
```
evanmcc, line 11: shouldn't this just point to master when merged to master?

PR #72, opened 2013-10-01
Add total_memory parameter and default value

engelsanchez:
This will pass a new open parameter to leveldb: total_memory. If the
parameter is not given to the eleveldb:open call, and if memsup is
running, we'll get it from memsup and pass it down.  Otherwise nothing
is passed to leveldb.

/cc @matthewvon Let me know if you need any tweaks


matthewvon:
code was manually merged into mv-flexcache branch.

PR #73, opened 2013-10-03
Mv flexcache

matthewvon:
This is the eleveldb side of mv-flexcache branch (already reviewed) in basho/leveldb.  The branch creates 3 config atoms:  total_memory, total_leveldb_mem, and total_leveldb_mem_percent.  Engel wrote code for total_memory.  The value is populated upon each database open for potential use with total_leveldb_mem_percent.  Any configuration value supplied by the user for total_leveldb_mem or total_leveldb_mem_percent is placed in the leveldb::Options.total_leveldb_mem structure member.  If between 0 and 100, it is assumed later to be a percent and applied against total_memory to achieve a byte value.

### DIFF
--- a/c_src/build_deps.sh
+++ b/c_src/build_deps.sh
@@ -8,7 +8,7 @@ if [ `uname -s` = 'SunOS' -a "${POSIX_SHELL}" != "true" ]; then
 fi
 unset POSIX_SHELL # clear it so if we invoke other scripts, they run as ksh as well
 
-LEVELDB_VSN="2.0.0pre1"
+LEVELDB_VSN="mv-flexcache"
 
 SNAPPY_VSN="1.0.4"
 
--- a/c_src/build_deps.sh
+++ b/c_src/build_deps.sh
@@ -8,7 +8,7 @@ if [ `uname -s` = 'SunOS' -a "${POSIX_SHELL}" != "true" ]; then
 fi
 unset POSIX_SHELL # clear it so if we invoke other scripts, they run as ksh as well
 
-LEVELDB_VSN="mv-flexcache"
+LEVELDB_VSN="b4f590fd44e865ef6fac0c7886ea324c3534b846"
 
 SNAPPY_VSN="1.0.4"
 
--- a/c_src/eleveldb.cc
+++ b/c_src/eleveldb.cc
@@ -387,8 +387,8 @@ async_open(
     // convert total_leveldb_mem to byte count if it arrived as percent
     //  This happens now because there is no guarantee as to when the total_memory
     //  value would be read relative to total_leveldb_mem_percent in the option fold
-    if (0 < opts.total_leveldb_mem && opts.total_leveldb_mem<=100)
-        opts.total_leveldb_mem=(opts.total_leveldb_mem * gCurrentTotalMemory)/100;
+    if (0 < opts->total_leveldb_mem && opts->total_leveldb_mem<=100)
+        opts->total_leveldb_mem=(opts->total_leveldb_mem * gCurrentTotalMemory)/100;
 
     eleveldb::WorkTask *work_item = new eleveldb::OpenTask(env, caller_ref,
                                                               db_name, opts);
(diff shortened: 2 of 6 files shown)


c_src/build_deps.sh
```diff
@@ -8,7 +8,7 @@ if [ `uname -s` = 'SunOS' -a "${POSIX_SHELL}" != "true" ]; then
 fi
 unset POSIX_SHELL # clear it so if we invoke other scripts, they run as ksh as well
 
-LEVELDB_VSN="2.0.0pre1"
+LEVELDB_VSN="mv-flexcache"
```
jrwest: this should be removed before merge right?

c_src/eleveldb.cc
```diff
@@ -249,6 +227,56 @@ ERL_NIF_TERM parse_open_option(ErlNifEnv* env, ERL_NIF_TERM item, leveldb::Optio
...
         }
+        else if (option[0] == eleveldb::ATOM_TOTAL_MEMORY)
+        {
+            // NOTE: uint64_t memory_sz and enif_get_uint64() do NOT compile
+            // correctly on some platforms.  Why?  because it's Erlang.
+            unsigned long memory_sz;
+            if (enif_get_ulong(env, option[1], &memory_sz))
+            {
+                // ignoring memory size below 2G, going with defaults
+                if (1024*1024*1024L < memory_sz)
```
jrwest, line 240: is the comment wrong or should of those be 2048?

c_src/build_deps.sh
matthewvon: Yes ... but Jared typically changes it.  I will also before check-in.

jrwest:
re: cuttlefish schema (cc/ @joedevivo). some of this may have been discussed already, sorry if I'm making you repeat yourself.
1. Are we planning to standardize on `on/off` vs. `true/false` for boolean values? This schema uses bolth (bloom filter and sync). Also, the `on/off` translation seems to be pretty common, what do you think about cuttlefish having a "stdlib" of translations for things like this?
2. The multibackend config being in this repo seems wrong -- there are references to `riak_kv` which depends on this repo.

c_src/eleveldb.cc
matthewvon, line 240: Bad comment ... will update.  The idea is that Erlang/Riak need about 1G to run without regard to leveldb.  If the total machine is NOT greater than 1G ... give up on automatic sizing.

src/eleveldb.erl
```diff
@@ -290,6 +295,27 @@ validate_options(Type, Opts) ->
...
 %% ===================================================================
+%% @doc Appends default open arguments that are better figured out
+%% in the Erlang side of things. Most get a default down in leveldb
+%% code. Currently only system total memory reported by memsup,
+%% if available.
+add_open_defaults(Opts) ->
+    case not proplists:is_defined(total_memory, Opts)
+        andalso is_pid(whereis(memsup)) of
+        true ->
+            case proplists:get_value(system_total_memory,
```
jrwest, line 306: I'm pretty sure I've deduced the reason but @engelsanchez do you think it would be worth commenting as to why we use `system_total_memory` instead of `total_memory`?

test/cacheleak.erl
```diff
@@ -30,7 +30,7 @@ cacheleak_test_() ->
                               [] = os:cmd("rm -rf /tmp/eleveldb.cacheleak.test"),
                               Blobs = [{<<I:128/unsigned>>, compressible_bytes(10240)} ||
                                           I <- lists:seq(1, 10000)],
-                              cacheleak_loop(10, Blobs, 350000)
+                              cacheleak_loop(10, Blobs, 400000)
```
jrwest: I had to raise this to `500000` to get it to pass. It passed on `master` w/o change fwiw.

src/eleveldb.erl
matthewvon, line 306: Fritchie sent us looking here:  http://www.snookles.com/erlang-docs/R15B03/lib/os_mon-2.2.10/doc/html/memsup.html#get_system_memory_data-0

I believe the idea was that total_system_memory was better supported across platforms.

c_src/build_deps.sh
matthewvon: going to patch this line last, with SHA of leveldb merge to master.

jrwest:
Built and ran w/ riak against `verify_build_cluster`. Options are clearly getting passed because before checking out `mv-flexcache` in basho/riak no values were getting passed in and I would hit the 10Mbyte lower limit. Checking out that branch in basho/riak and basho/riak_core and rebuilding fixed it (so the options clearly take). All eleveldb tests pass as well.

I do have reserves about some of the way we are doing config for the multi-backend but that is a caremad for another day. 

+1. merge.

PR #75, opened 2013-10-19
Mv flexcache4

matthewvon:
Create load library level options so that Riak configs.  This helps backends that do NOT use leveldb but need the memory limits to control AAE.  Also handles race condition where AAE loads AFTER leveldb backend and screws up memory limits (and made Matthew's machines go to swap).


vinoski:
+1 to merge, looks good.

PR #76, opened 2013-10-23
add support for OS X Mavericks

vinoski:
Enable eleveldb to run on OS X Mavericks. Pick up a new leveldb commit, and
add new Mavericks-specific switches to rebar.config.

Tested successfully on Mavericks, Mountain Lion, and Lion.

Requires leveldb changes too, see https://github.com/basho/leveldb/pull/104 .


joedevivo:
+1 This does, in fact, support Mavericks. Would love a +1 from @matthewvon as well.

cmeiklejohn:
@matthewvon wasn't this supposed to make it into pre5?  I thought we wanted this to be runnable.

matthewvon:
To make a long story short … leveldb sv-mavericks-port was merged.  This never hit my radar and was never considered.

PR #77, opened 2013-11-12
fixed dialyzer errors in async_iterator_move

Licenser:
Hi I fixed two problems with the specs of async_iterator_move. That said please double check of this new specs are correct, they make sense from the way the functions is called and from looking at the C-Code it also seems to be the proper return but I'm not that deep into the levedb C-code that I feel 100% confident about reading it correctly.


matthewvon:
I have sent out a review request.  Dialyzer and Erlang are not my strengths.

andrewjstone:
This fixes an improper spec error from dialyzer and looks correct from my reading of the C code as well. :+1:

PR #78, opened 2013-11-13
Batch iterators - improved performance.

Lol4t0:
Grouping iterator movement operations allows to improve performance, so 'fold' runs ~10 times faster on my machine, than on original eleveldb when using batch_size ~1000 .

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ c_src/system
 *~
 .project
 .settings
+/eleveldb_c/
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@ c_src/leveldb/
 c_src/snappy*/
 c_src/system
 *~
+.project
+.settings
--- a/c_src/eleveldb.cc
+++ b/c_src/eleveldb.cc
@@ -718,6 +718,9 @@ async_iterator_move(
         submit_new_request=true;
         ret_term = enif_make_copy(env, itr_ptr->m_Snapshot->itr_ref);
 
+        //turn off prefetch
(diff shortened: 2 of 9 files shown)


matthewvon:
Are you seeing 10x improvement under Riak or some other application?

Lol4t0:
First, it is fold implementation that shows 10x performance gain, not the whole application. You can get approximate results with benchmarks provided in fold_test.erl 

Then, you should support batch iterators (i.e. set appropriate batch size) to make your app run faster. With batch size <10 you can even lose some performance.

matthewvon:
An interested, internal Basho developer reviewed this code for possible use in a new feature.  He was unable to validate the 10x claims using an Apple laptop with SSD drives.  Without an internal Basho sponsor the code has no future.

The code has structure similar to something written previously by another Basho developer.  The previous code did not have the expected throughput gain either.

You would need to supply example code and test output for this PR to remain open.  Otherwise, we will close it without adding to the main code base after July 2015.

Lol4t0:
There is indeed a test already in the code. Look through `batch_read_jump` in `test/iterators.erl`. The original claim is based on that code results.

To get sufficient performance gain
- Test array should be large enough and batch size should be large enough
- Iterating should be implemented carefully. While using the same code structure you can kill performance with the single extra synchronization with mutex or smth like that. I've tried to do my best to implement wait free approach while implementing that feature.

russelldb:
I used a db of 100k elements, and a batch size of 10,100, 1000, 10000. I'll do further testing with more fine grained measurement this week. I'd rather @matthewvon didn't close this just yet, I have an interest, but need to verify the results first.

PR #81, opened 2013-12-04
mv-fadvise-control

matthewvon:
Changes discussed here https://github.com/basho/leveldb/wiki/mv-fadvise-control

### DIFF
--- a/c_src/eleveldb.cc
+++ b/c_src/eleveldb.cc
@@ -205,7 +205,7 @@ ERL_NIF_TERM parse_init_option(ErlNifEnv* env, ERL_NIF_TERM item, EleveldbOption
 {
     int arity;
     const ERL_NIF_TERM* option;
-    if (enif_get_tuple(env, item, &arity, &option))
+    if (enif_get_tuple(env, item, &arity, &option) && 2==arity)
     {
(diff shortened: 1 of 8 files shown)


c_src/eleveldb.cc
```diff
@@ -156,23 +159,74 @@ static ERL_NIF_TERM slice_to_binary(ErlNifEnv* env, leveldb::Slice s)
...
 };
 
 
+ERL_NIF_TERM parse_init_option(ErlNifEnv* env, ERL_NIF_TERM item, EleveldbOptions& opts)
+{
+    int arity;
+    const ERL_NIF_TERM* option;
+    if (enif_get_tuple(env, item, &arity, &option))
+    {
+        if (option[0] == eleveldb::ATOM_WRITE_THREADS)
```
engelsanchez, line 210: We don't really check for arity here, so we could crash with a badly formed tuple of size 1 or zero. It would be nice to check.

engelsanchez:
+1  I've reviewed the code and tested under Linux using strace to see actual fadvise usage (in Linux it uses fadvise64). Works as advertised.

engelsanchez:
This PR needs to be retargeted to point to 1.4 though. Don't forget that!

matthewvon:
Manually merged to 1.4 branch

PR #86, opened 2013-12-12
Mv fadvise control 2.0

matthewvon:
This is the customer requested feature first merged into 1.4.4. Copied here for 2.0 integration.  This PR also adds three other leveldb option variables that are too simple to justify an independent PR.

Notes on the feature are here: https://github.com/basho/leveldb/wiki/mv-fadvise-control


andrewjstone:
builds, tests pass, code looks good. :+1:

PR #87, opened 2013-12-12
Schema updates for new cuttlefish API. unit test.

joedevivo:


Makefile
```diff
@@ -1,9 +1,16 @@
...
+ifeq ($(REBAR_BIN),)
+REBAR_BIN = ./rebar
+endif
 
 all: compile
 
 get-deps:
 	./c_src/build_deps.sh get-deps
 
+deps:
```
seancribbs, line 11: Shouldn't `compile` depend on this target? Also it should be declared as `.PHONY` at the top, since `deps` is actually a directory in the project.

joedevivo, line 11: I did it the way it was done in bitcask, for better or worse I guess.

seancribbs, line 11: Ok, but means `make compile` doesn't work out-of-box.

seancribbs:
Other than the Makefile issue, I'm :+1: on this.

seancribbs:
:+1:

PR #88, opened 2013-12-19
Mv iterator refresh

matthewvon:
Special case code:  when an iterator is held for a long, long time AND its database (vnode) is under heavy write load, the snapshot held by the iterator can cause that database thrash.  If this iterator is ONLY processing NEXT operations, this branch will close and reopen the iterator every 5 minutes.  Intended primarily for internal Riak operations such as AAE.

Notes here:  https://github.com/basho/leveldb/wiki/mv-iterator-refresh

### DIFF
--- a/c_src/refobjects.h
+++ b/c_src/refobjects.h
@@ -213,59 +213,6 @@ private:
 };  // class DbObject
 
 
-
-/****
- ** Note Dec 18, 2013, matthewv:  LevelSnapshotWrapper and LevelIteratorWrapper
- **  are left over from a design that did not perform as well as hoped.  There
- **  so no reason why they could not be rolled into ItrObject.
- ***/
-
-/**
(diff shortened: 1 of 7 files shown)


c_src/refobjects.h
```diff
@@ -314,15 +304,19 @@ class ItrObject : public ErlRefObject
 {
 public:
     ReferencePtr<LevelIteratorWrapper> m_Iter;
-    ReferencePtr<LevelSnapshotWrapper> m_Snapshot;
+//    ReferencePtr<LevelSnapshotWrapper> m_Snapshot;
```
andrewjstone, line 307: This line should be deleted

andrewjstone:
- [x] Code looks great 
- [x] Conforms with description in prose
- [x] tests pass

:+1:

PR #91, opened 2014-01-09
restore AAE exemption to writeblock size accounting in write throttle

jaredmorrow:
Restore AAE exemption to writeblock size accounting in write throttle, BUT leave the accounting for 2i writeblocks.  Required back-porting of Riak 2.0 is_internal_db flag

### DIFF
--- a/c_src/build_deps.sh
+++ b/c_src/build_deps.sh
@@ -8,7 +8,7 @@ if [ `uname -s` = 'SunOS' -a "${POSIX_SHELL}" != "true" ]; then
 fi
 unset POSIX_SHELL # clear it so if we invoke other scripts, they run as ksh as well
 
-LEVELDB_VSN="1.4.4"
+LEVELDB_VSN="mv-aae-writeblock"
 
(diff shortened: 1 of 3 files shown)


c_src/build_deps.sh
```diff
@@ -8,7 +8,7 @@ if [ `uname -s` = 'SunOS' -a "${POSIX_SHELL}" != "true" ]; then
 fi
 unset POSIX_SHELL # clear it so if we invoke other scripts, they run as ksh as well
 
-LEVELDB_VSN="1.4.4"
+LEVELDB_VSN="mv-aae-writeblock"
```
reiddraper, line 11: presumably @jaredmorrow updates this once the change is tagged in basho/leveldb?

reiddraper:
Is a clear backport of 3493e4fecfb07b67e727a4b2fa09494a7eedf610, +1. All 18 tests pass.

c_src/build_deps.sh
jaredmorrow, line 11: I do indeed update that for release.

PR #104, opened 2014-03-13
correct prefetch race condition.

matthewvon:
This PR corrects a race condition in the iterator prefetch.  The race was found using 5 node cluster with bitcask backend.  AAE's use of leveldb triggered the race and resulted in segfault.  All is good now.


andrewjstone:
:+1: Great verbal walkthrough with @matthewvon

PR #105, opened 2014-03-13
Fix dialyzer and xref errors

Vagabond:
Also update tools.mk to 0.5.3


andrewjstone:
Looks good. Dialyzer passed. :+1:

PR #80, opened 2014-03-27
multibackend part of .schema belongs into riak_kv

joedevivo:
Hi I think the multibackend part of the .schema file does not belong into the eleveldb.schema file. It's specific to riak_kv and not to this library so I think it would belong there.

This can of cause be argued given eleveldb was developed for riak_kv but still for library users it would make more sense to keep those separated.

An alternative would be having two .schema files one eleveldb.schema and one eleveldb_multi.schema

### DIFF
--- a/priv/eleveldb.schema
+++ b/priv/eleveldb.schema
@@ -7,12 +7,6 @@
   hidden
 ]}.
 
-%% @see leveldb.data_root
-{mapping, "multi_backend.$name.leveldb.data_root", "riak_kv.multi_backend", [
-  {datatype, directory},
-  hidden
-]}.
-
 %% @doc This parameter defines the percentage of total server memory
 %% to assign to LevelDB. LevelDB will dynamically adjust its internal
 %% cache sizes to stay within this size.  The memory size can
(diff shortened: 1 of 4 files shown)


jrwest:
funny you should ask, it just came back up today cc/ @joedevivo

matthewvon:
Part of the answer is simple.  If you move the parallel settings of multibackend to another repo / location, they will not be maintained.  Either parallel has to be eliminated, or kept here.

jrwest:
imo, the fundamental problem is that the settings need to be duplicated, rather than only being defined once and "included" into the multi-backend as "subconfigs". That's not something thats supported right now though, so they should live in the same repo.

joedevivo:
I'm working on a bug in the bitcask multi_backend schema, in which I'll try to accommodate this same issue. If we like the solution, I'll port it to leveldatabass

joedevivo:
Check out how I handled it in basho/bitcask#157

joedevivo:
The above is the application of the solution used in basho/bitcask#157

matthewvon:
is schema_tests verifying that all items in eleveldb schema are in eleveldb_multi schema?  my erlang sucks.

joedevivo:
@matthewvon yup, and that they're properly overridden with multi versions of the defaults if necessary

jrwest:
+1 merge. builds and tests pass. Like https://github.com/basho/bitcask/pull/157 Riak & EE will need to have the new schema's incorporated into the build for this all to work.

PR #113, opened 2014-04-25
avoid badarg exception when closing db_refs being closed

vinoski:
DbObject::RetrieveDbObject() returns NULL for 2 cases: when the db_ref NIF
resource can't be retrieved, and when the "close requested" flag is already
set. But eleveldb_close() treats both NULL return cases as an error and
raises the badarg exception for both.

Modify DbObject::RetrieveDbObject() to add a defaulted third argument, a
pointer to bool that defaults to NULL. If non-NULL,
DbObject::RetrieveDbObject() sets the bool to true if the db_ref NIF
resource is successfully retrieved, false otherwise. This allows
eleveldb_close() to pass that bool argument and then use it to distinguish
between the two NULL return cases from DbObject::RetrieveDbObject().

### DIFF
--- a/c_src/eleveldb.cc
+++ b/c_src/eleveldb.cc
@@ -947,10 +947,11 @@ eleveldb_close(
 {
     eleveldb::DbObject * db_ptr;
     ERL_NIF_TERM ret_term;
+    bool term_ok=false;
 
     ret_term=eleveldb::ATOM_OK;
 
-    db_ptr=eleveldb::DbObject::RetrieveDbObject(env, argv[0]);
+    db_ptr=eleveldb::DbObject::RetrieveDbObject(env, argv[0], &term_ok);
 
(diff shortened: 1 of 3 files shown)


reiddraper:
+1. I've run eleveldb eunit/eqc 5 times now with no problems. I've run KV once (and offline @vinoski mentioned he ran it overnight). I'll continue running KV eunit to be safe. But let's go ahead and merge.

reiddraper:
The merge-commit [test failed](http://buildbot.bos1/builders/test-eleveldb/builds/301/steps/shell_3/logs/stdio). @vinoski and I are looking into it. It's failing with `process killed by signal 11`.

reiddraper:
I [forced a rebuild](http://buildbot.bos1/builders/test-eleveldb/builds/302) and the test still failed.

PR #117, opened 2014-06-19
Mv iter close fix

matthewvon:
Correct race condition in asynchronous close logic that leads to segfaults.  The bad code was merged to develop via the mv-tuning7 branch.

Detail discusssion is here:  https://github.com/basho/leveldb/wiki/mv-iter-close-fix

### DIFF
--- a/c_src/eleveldb.cc
+++ b/c_src/eleveldb.cc
@@ -772,7 +772,7 @@ async_iterator_move(
 
     itr_ptr.assign(ItrObject::RetrieveItrObject(env, itr_handle_ref));
 
-    if(NULL==itr_ptr.get())
+    if(NULL==itr_ptr.get() || 0!=itr_ptr->m_CloseRequested)
         return enif_make_badarg(env);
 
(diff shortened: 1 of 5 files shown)


vinoski:
+1

c_src/eleveldb.cc
```diff
@@ -954,7 +954,8 @@ async_close(
     // verify that Erlang has not called DbObjectResourceCleanup
     //  already (that would be bad)
     if (NULL!=db_ptr->m_Db
-        && compare_and_swap(db_ptr->m_ErlangThisPtr, db_ptr.get(), (DbObject *)NULL))
+//        && compare_and_swap(db_ptr->m_ErlangThisPtr, db_ptr.get(), (DbObject *)NULL))
```
reiddraper, line 957: any reason to leave commented code in?

matthewvon, line 957: Awaiting next round of changes to clean up.  With this bug, never completely sure if you are finished.

PR #119, opened 2014-07-10
Must wait until complete close of iterator finishes. 

matthewvon:
A previous hack allowing iterator close to return immediately created race condition that would lead to Erlang heap corruption/crash.  This could be an Erlang heap issue (unlikely) or an uncovered race condition elsewhere in the eleveldb NIF (more likely).  For now, the crash goes away when the iterator close call properly awaits a message saying leveldb's close has completed.


vinoski:
+1. Ran tests under valgrind a number of times with zero failures.

PR #122, opened 2014-07-25
vsn in app.src should be a string

carlhoerberg:
http://www.erlang.org/doc/design_principles/applications.html#id73836


seancribbs:
Rebar replaces the 'git' atom with the output of `git-describe` when generating the `.app` file. This is not needed.

PR #126, opened 2014-08-21
prior mv-complete-itr-close branch was a guess. this is now a fix to roo...

matthewvon:
...t cause.  (steve v. isolated via valgrind then verified this fix)


matthewvon:
steve v. already verified elsewhere

PR #136, opened 2015-03-14
- made code compile even if size_t not being long integer

franzhaas:
I changed the usage of size_t to unsigned long, and changed the 8G constant to LL.


seancribbs:
Are you compiling on a 32-bit platform?

franzhaas:
Yes I am.

seancribbs:
Sorry, we explicitly don't support 32-bit. Our fork of leveldb uses various x86-64 features. Even if this makes eleveldb compile on your platform, the resulting linked library will probably be broken.

PR #137, opened 2015-03-30
Async Put Support

jonmeredith:
Expose the internal async put mechanism - callers can pass in a context to receive as a message on the calling process.


jonmeredith:
+1 d946c610ebd1aff5780293cc985eef611b8cd747

fadushin:
@borshop merge

PR #140, opened 2015-06-14
Remove lager parse transform

ahf:
Whilst debugging something where I disabled the cuttlefish dependency, I noticed that lager applied the lager parse transform without actually using lager anywhere. This patch removes the `lager_transform` from `erl_opts`.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -6,7 +6,7 @@
 
 {port_sources, ["c_src/*.cc"]}.
 
-{erl_opts, [warnings_as_errors, {parse_transform, lager_transform}, debug_info]}.
+{erl_opts, [warnings_as_errors, debug_info]}.
 
 {deps, [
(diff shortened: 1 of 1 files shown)


jonmeredith:
Is this a code hygiene issue for you, or were you trying to use eleveldb in a project without lager?
(seems unlikely given the eleveldb -> cfish -> lager dep chain)

ahf:
I'm not sure I'd use the term code hygiene here.

It's unclean to have a parse transform applied from something that isn't an explicit dependency IMO, and yes, I'm using eleveldb in an environment where we are not using cuttlefish and thus doesn't fetch the implicit lager dependency (which works just fine as well). There's no calls to lager:\* inside of eleveldb itself.

tsloughter:
Yea, it is just not right, so why keep it?

jonmeredith:
Agreed.

jonmeredith:
Applied to 2.0 and 2.1 branches too.

PR #142, opened 2015-07-02
Move DestroyDB call from erlang thread to background thread.

matthewvon:


src/eleveldb.erl
```diff
@@ -250,13 +250,16 @@ status(Ref, Key) ->
...
+
 -spec destroy(string(), open_options()) -> ok | {error, any()}.
 destroy(Name, Opts) ->
-    eleveldb_bump:big(),
-    destroy_int(Name, Opts).
-
-destroy_int(_Name, _Opts) ->
-    erlang:nif_error({erlang, not_loaded}).
+    CallerRef = make_ref(),
+    Opts2 = add_open_defaults(Opts),
```
engelsanchez, line 260: I imagine you don't need the default open opts here, but was just a cut n' paste addition?

matthewvon, line 260: The DestroyDB call does require leveldb::Options structure as a parameter (that is same as open_options here).

src/eleveldb.erl
```diff
@@ -250,13 +250,16 @@ status(Ref, Key) ->
...
+    async_destroy(CallerRef, Name, Opts2),
+    ?WAIT_FOR_REPLY(CallerRef).
```
engelsanchez, line 262: @matthewvon here is where we need to change the code to match the `{ok}` the task is returning to the `ok` it was expecting before:

```
{ok} = ?WAIT_FOR_REPLY(CallerRef),
ok.
```

Wrapping a single value is silly, but doing that allows us not to mess with existing code in https://github.com/basho/eleveldb/blob/mv-async-destroy/c_src/work_result.hpp#L62 that wraps a single result in a tuple.

c_src/workitems.cc
```diff
@@ -361,6 +361,32 @@ MoveTask::recycle()
...
+}   // DestroyTask::DestroyTask
+
+
+work_result
+DestroyTask::operator()()
+{
+    leveldb::Status status = leveldb::DestroyDB(db_name, *open_options);
+
+    if(!status.ok())
+        return error_tuple(local_env(), ATOM_ERROR_DB_OPEN, status);
```
engelsanchez: Cut n' paste error: Should be ATOM_ERROR_DB_DESTROY

src/eleveldb.erl
matthewvon, line 262: Fixed this by changing return from DestroyTask::operator() in workitems.cc instead of eleveldb.erl.

engelsanchez, line 262: For posterity: @matthewvon fixed the result by using a different work_result constructor in `DestroyTask::operator()`, which is way smarter.

engelsanchez:
+1 1d5c32a

Code looks good now, tests pass. I confirmed the interface is now the same after the last fix. S'all good.

PR #144, opened 2015-08-13
unsigned long for memory_sz so it compiles on scaleway C1 / arm

bougueil:


NoKriK:
Amazing, just tested it and it works!

Issue #141, opened 2015-06-28
Thread pool synchronizing problem? [JIRA: RIAK-1912]


agilesoftware:
in, `submit(eleveldb_thread_pool::submit)` :

```
 else if (!FindWaitingThread(item))
         {
             // no waiting threads, put on backlog queue
             lock();
             eleveldb::inc_and_fetch(&work_queue_atomic);
             work_queue.push_back(item);
             unlock();

             // to address race condition, thread might be waiting now
             FindWaitingThread(NULL);

             perf()->Inc(leveldb::ePerfElevelQueued);
             ret_flag=true;
         }   // if 
```

It first check if there's an waiting thread to submit directly(and pthread_cond_broadcast in `FindWaitingThread`), or it will add `item` to work_queue.

On the other hand, in eleveldb_write_thread_worker(threading.cc), when a working thread cannot find a work item, it will wait on its condition variable.

```
            pthread_mutex_lock(&tdata.m_Mutex);
            tdata.m_DirectWork=NULL; // safety

            // only wait if we are really sure no work pending
            if (0==h.work_queue_atomic)
        {
                // yes, thread going to wait. set available now.
            tdata.m_Available=1;
                pthread_cond_wait(&tdata.m_Condition, &tdata.m_Mutex);
        }    // if

            tdata.m_Available=0;    // safety
            submission=(eleveldb::WorkTask *)tdata.m_DirectWork; // NULL is valid
            tdata.m_DirectWork=NULL;// safety

            pthread_mutex_unlock(&tdata.m_Mutex);
```

after checking the working queue is empty(work_queue_atomic==0), it sets its available to 1. 

The problem is, if an working thread has checked that the working queue is empty , and before it's setting available, `submit` maybe cannot find a waiting thread, add an item to queue. At this moment, the working thread starts waiting, then this working thread will wait forever.(The second `FindWaitingThread(NULL)` doesn't help, it also maybe executed in this period). In fact, if the working thread has setting its available, and before waiting on  its condition, and `submit`  set its `direct_work`, broadcast the condition variable, and the working thread wait, it will wait forever too.

Basho-JIRA:
Is this a question based upon code review or an experienced condition? 

**_[posted via JIRA by Matthew Von-Maszewski]**_

Basho-JIRA:
This condition was addressed in the leveldb version of this code via an extra thread that is based upon sem_post().  Said extra thread logic could be back ported. 

[...]

PR #147, opened 2015-09-11
pp hot threads backport

paulplace:
This branch backports the hot_threads class from leveldb to eleveldb.

The corresponding pull request for leveldb is https://github.com/basho/leveldb/pull/171.

The thread pool in eleveldb, eleveldb_thread_pool, contains a race condition in the job hand-off code that is the subject of at least one bug report (https://github.com/basho/eleveldb/issues/141). This race condition was addressed in the hot_threads code added to leveldb (https://github.com/basho/leveldb/wiki/mv-hot-threads).

This branch removes eleveldb_thread_pool, instead using the hot_threads code in leveldb.

Other work done in this branch is to remove duplicated code in favor of using the common code in leveldb:
- detail.hpp was removed; the atomic operations in include/leveldb/atomics.h are now used
- mutex.h was removed; the Mutex and MutexLock classes in leveldb are now used
- replaced all pthread_mutex_t and pthread_cond_t variables with the corresponding wrapper classes in leveldb
- the RefObject and ErlRefObject classes now use the RefObjectBase class in leveldb instead of performing their own reference counting

Caveats:
- the change in build_deps.sh ties this branch to the corresponding branch in leveldb
-  the manifest constant LEVELDB_PLATFORM_POSIX is hard-coded in the files that now use code from leveldb's porting layer


matthewvon:
+1 is ready to merge ... local OS X tests now pass.  Build on Basho auto build now succeeds.  We are READY!!!

PR #148, opened 2015-09-24
Feature/ignore nonmsgpack

erikleitch:
This branch includes two functional changes, and one change to a unit test file:
- workitems.cc has been modified so that records that are non-msgpack encoded will be ignored during a streaming fold.  This handles the case where msgpack-encoded and non-magpack-encoded records are interleaved in the same leveldb file
- eleveldb:do_streaming_fold() has been modified to log errors thrown up from the C++ layer during filtering
- I've added test function r() to test/sut.erl as another example of decoding TS values returned from eleveldb, and of correct filter formatting


c_src/extractor.cc
```diff
@@ -338,6 +338,18 @@ DataType::Type Extractor::cTypeOf(ErlNifEnv* env, ERL_NIF_TERM tuple, bool throw
  * Given the start of a key data binary, return the start and size of the
  * contents portion of an encoded riak object
  */
+bool Extractor::riakObjectContentsCanBeParsed(const char* data, size_t size)
+{
+    const char* ptr = data;
+    ptr++;
+    unsigned char vers     = (*ptr++);
```
hmmr, line 345: There is a magic number at offset 0: https://github.com/basho/riak_kv/blob/develop/src/riak_object.erl#L79. Perhaps you should be checking it as well?

c_src/workitems.cc
```diff
@@ -767,17 +769,50 @@ work_result RangeScanTask::operator()()
...
-		parse_range_filter_opts(options_.env_, 
-                                        options_.rangeFilterSpec_, *(options_.extractor_),
-                                        throwIfBadFilter);
-	    }
+                //------------------------------------------------------------
+                // Only attempt to parse the data if it is correctly
+                // formatted, otherwise we would throw an error. 
+                //------------------------------------------------------------
+
+                if(options_.extractor_->riakObjectContentsCanBeParsed(value.data(), value.size())) {
```
hmmr, line 783: If this condition is almost invariably be true (all data are time series data and hence satisfying the condition), then it would make sense to signal this to gcc by wrapping in it the `likely` macro.

test/sut.erl
```diff
@@ -854,8 +859,27 @@ badEncodingOptions_test() ->
...
 
+    Cond1 = {'=', {field, <<"user">>, binary},    {const, <<"user_1">>}},
+    Cond2 = {'<', {field, <<"time">>, timestamp}, {const, 11000000}},
+    Cond3 = {'>', {field, <<"time">>, timestamp}, {const,  9990000}},
+    Filter = {'and_', Cond1, {'and_', Cond2, Cond3}},
+    Opts=[{fold_method, streaming},
+	  {range_filter, Filter},
+	  {encoding, msgpack}],
+    io:format("Filter = ~p~n", [lists:flatten([Filter])]),
+    TableName = "/Users/eml/projects/riak/riak_end_to_end_timeseries/dev/dev1/data/leveldb/" ++ Table,
```
hmmr, line 1078: I'm afraid make test will surely not pass on my machine :}

russelldb:
I don't understand _why_ these types would ever be interleaved. Wouldn't it make folds of unknown expense if TS data is not stored together?

Riak Object data is stored under a key that is `sext:encode({o, Bucket, Key})` and indexes are `sext:encode({i, Bucket, Field, Term, Key})`

Can't you ensure that TS objects have a key that ensures they are "clumped" together?

russelldb:
My point is, why skip over? It seems that a non TS series datum in a fold is an indication of a serious error, no?

hmmr:
+1

gordonguthrie:
JIRA raised to address Russell's key grouping point
https://bashoeng.atlassian.net/browse/RTS-345

gordonguthrie:
@borshop merge

gordonguthrie:
+1 069f6d4

PR #149, opened 2015-09-25
Reverted first_key/last_key fold options to start_key/end_key.  This …

erikleitch:
…change was requested during dev of the eleveldb branch, but breaks riak_kv test code (and probably others) that still references start_key/end_key, and also creates inconsistent naming convention wrt other options that still reference start/end.  So we're changing it back.  Again.


gordonguthrie:
+1 94e3aaa

PR #150, opened 2015-09-30
Moved ownership of extractor_ and range_filter_ from RangeScanOptions…

erikleitch:
… to RangeScanTask to simplify memory management and fix the memory leak introduced by extractor_ not being deleted in ~RangeScanOption()

### DIFF
--- a/c_src/workitems.cc
+++ b/c_src/workitems.cc
@@ -487,19 +487,40 @@ RangeScanTask::RangeScanTask(ErlNifEnv * caller_env,
                              RangeScanOptions & options,
                              SyncObject * sync_obj)
 : WorkTask(caller_env, caller_ref, db_handle),
-    options_(options),
-    start_key_(start_key),
-    has_end_key_(bool(end_key)),
-    sync_obj_(sync_obj)
+  options_(options),
+  start_key_(start_key),
+  has_end_key_(bool(end_key)),
+  sync_obj_(sync_obj),
+  range_filter_(0),
+  extractor_(0)
 {
     if (end_key) {
         end_key_ = *end_key;
     }
+
+    if(options_.useRangeFilter_) {
+        if(options_.encodingType_ == Encoding::MSGPACK)
+            extractor_ = new ExtractorMsgpack();
+        else {
+            ThrowRuntimeError("An invalid object encoding was specified");
+        }
+    }
+
     sync_obj_->RefInc();
 }
 
 RangeScanTask::~RangeScanTask()
 {
+    if(range_filter_) {
+        delete range_filter_;
+        range_filter_ = 0;
+    }
+    
+    if(extractor_) {
+        delete extractor_;
+        extractor_ = 0;
+    }
+
     sync_obj_->RefDec();
 }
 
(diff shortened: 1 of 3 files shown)


erikleitch:
This is an internal change to eleveldb (no repercussion for erlang/elevldb interface) to fix a small memory leak I've identified in the C++ layer.

erikleitch:
Some additional context: this was a problem that originated with the original quick-n-dirty TS prototype: in the protoype/timeseries-range-scan-filter branch, these pointers belonged to RangeScanOptions, and were allocated but never deleted anywhere.  Because RangeScanOptions gets passed to RangeScanTask and copied into that object,  an implicit copy constructor for RangeScanOptions was getting called that duplicated those pointers, making it unsafe to delete those members in ~RangeScanOptions().  The purpose of this PR is to:

1) move responsibility for allocation and deletion of these pointers to RangeScanTask, which is the object that actually uses them

2) by changing ownership, ensure that the pointers are allocated once and never copied

3) explicitly free these objects in ~RangeScanTask(), ensuring that the memory for these objects is released

c_src/workitems.cc
```diff
@@ -487,19 +487,40 @@ RangeScanTask::RangeScanTask(ErlNifEnv * caller_env,
...
 {
     if (end_key) {
         end_key_ = *end_key;
     }
+
+    if(options_.useRangeFilter_) {
+        if(options_.encodingType_ == Encoding::MSGPACK)
+            extractor_ = new ExtractorMsgpack();
+        else {
+            ThrowRuntimeError("An invalid object encoding was specified");
```
matthewvon, line 505: streaming_start() in eleveldb.cc calls "new" operator on this object without try/catch.  Riak will crash if this throws.

c_src/workitems.h
```diff
@@ -629,6 +619,8 @@ namespace eleveldb {
         std::string end_key_;
         bool has_end_key_;
         SyncObject * sync_obj_;
```
matthewvon, line 621: Appears sync_obj_ is NOT optional.  Therefore it needs to be a reference, not a pointer.  If optional, pointer is not validated for NULL and could cause segfault.

c_src/workitems.cc
```diff
@@ -791,9 +812,9 @@ work_result RangeScanTask::operator()()
                     //------------------------------------------------------------
 
                     try {
-                        options_.range_filter_ = 
+                        range_filter_ = 
                             parse_range_filter_opts(options_.env_, 
-                                                    options_.rangeFilterSpec_, *(options_.extractor_),
+                                                    options_.rangeFilterSpec_, *(extractor_),
                                                     throwIfBadFilter);
                     } catch(std::runtime_error& err) {
```
matthewvon, line 827: This is going to throw.  I believe it is throwing to threading.cc notify_caller() which is called by eleveldb_write_thread_worker().  I do not see try/catch block anywhere in that call stack.

No, there is a surrounding try/catch I missed.

erikleitch:
Committed a few changes to feature/bugfix-extractorleak in response to Matthew's comments:

1) added protection around the RangeScanTask constructor, which can now in principle throw

2) added check for NULL SyncObj on construction (I don't want to rampage through that code further until I understand more about how that obj is used)

3) made threadpool sizes prime

matthewvon:
+1

Issue #152, opened 2015-10-01
Cuttlefish should be bumped to develop to compile on Erlang 18 [JIRA: RIAK-2212]


TBK145:
The rebar.config from the dependency Cuttlefish 2.0.1 is only for R16|17, however, this is fixed on the development branch, so the depency should be bumped to {branch: "develop"}.
https://github.com/basho/cuttlefish/blob/develop/rebar.config

PR #153, opened 2015-10-02
Bumped cuttlefish to 'develop' so it compiles on Erlang 18.

TBK145:
Issue here: https://github.com/basho/eleveldb/issues/152


bsparrow435:
Apologies for the delay, we're currently in the process of updating all Riak libraries to support OTP18/19 and will get this merged as part of that work. Again, thank you for your contribution.

PR #160, opened 2015-10-15
Increased both stream thread poool and normal thread pool to first pr…

erikleitch:
…ime > 64 (upper limit for typical single-machine cluster size)

When I merged the streaming folds code into branch end-to-end/timeseries, I added a second thread pool to eleveldb, so that large streaming folds wouldn't block normal put/get operations.  But I maintained the total number of threads from the original single pool.  After discussion with mvm, we agreed that both pools should be the same size, and large enough to acomodate typical large single-machine cluster sizes.

This PR is a one-line change to resize both thread pools to be the same size as the original thread pool (71).


erikleitch:
Well, not the first prime, but a prime with headroom :-)

matthewvon:
+1

PR #161, opened 2015-10-15
Modified msgpack extraction path to check for empty erlang list (null…

erikleitch:
… indicator from riak_kv).  Modified filter to return false when evaluated against a missing field value, instead of throw.  Modified unit tests to reflect the change in behavior.

Note: 

The filtering code was originally developed under the direction that missing fields should be an error.  This PR changes that behavior so that filters that reference missing fields evaluate to false, instead of throwing an error.

Summary of changes:

1) added code in CmpUtil to detect empty list encoding in msgpack
2) added code in extractor.cc to check for empty fields while decoding msgpack
3) modified all binary filter operators in filter.h to return false in evaluate() rather than throw if a field has no value
4) modified unit tests in sut.erl to reflect the new behavior


matthewvon:
Is .rebar/erlcinfo an intentional check-in ?  I do not know its function.

matthewvon:
+1 once .rebar/erlcinfo explained or removed.

erikleitch:
.rebar/erclinfo not intentional -- got checked in by accident at some point & I removed it from a later branch.   Will make sure it's removed from the merge

matthewvon:
When you commit the change to delete .rebar/erclinfo, would you include an edit to .gitignore that explicitly lists that file.   Might as well protect against the next time ...

Thank you.

erikleitch:
Yup.  Good idea.

PR #162, opened 2015-10-19
Feature/mainline leveldb

erikleitch:
PR to remove all C++11isms from both eleveldb and leveldb.  This allows timeseries eleveldb to use the current develop branch of leveldb.

This PR also removes all unused family/series constructs from both eleveldb and leveldb, and associated key comparator/translator code.    

Detailed notes:

**C++11isms**

1) build_deps.sh has been modified to use leveldb version tags/2.1.3
2) rebar.config has been modified to remove c++0x flags

**eleveldb.h/cc**

3) All 'family' versions of async_open/close/get/put have been removed from eleveldb.erl/cc

4) All TS-specific key translation/comparison has been removed.  Key comparison now defaults to normal leveldb bytewise comparison, and no translation of keys is done at all.

5) while the code in eleveldb.cc that had been modified to support the family construct has been reverted to elevledb/develop branch counterparts, where possible I have replaced the boilerplate code to submit tasks to the thread pool with a call to submit_to_thread_queue().

**workitems.h/cc**

6) The code in workitems.h had evolved to a messy state where half of the Task class functions were actually defined in workitems.h and half were defined in workitems.cc, more or less randomly.  While this PR essentially reverts things like the WriteTask/GetTask classes to their eleveldb/develop counterparts to remove the family behavior, I have migrated most of the workitem class functionality into workitems.cc, leaving workitems.h as a (mostly) clean interface description.  (so some code has moved around from workitems.h to workitems.cc, but should be directly comparable with the corresponding functions in elevledb/develop, for example)

**misc**

7) Removed a few test functions from eleveldb.erl/eleveldb.cc that shouldn't be part of production code

8) Added some comments to filter code (filter.h) and rearranged some unhappy-path code to improve type-checking efficiency


c_src/workitems.h
```diff
@@ -264,391 +220,275 @@ namespace eleveldb {
...
-            // so no process is there to consume the output.
-            void MarkConsumerDead();
-
-            bool IsConsumerDead() const;
-
-        private:
-            const uint32_t max_bytes_;
-            const uint32_t low_bytes_;
-            volatile uint32_t num_bytes_;
-            bool producer_sleeping_;
```
matthewvon, line 577: Should be volatile so that compiler knows it cannot optimize value into register.  Is used in tight loop around a condition variable to communicate state between threads.

matthewvon:
c_src/detail.hpp was updated for time series.  uint32_t operation looks wrong for Solaris.  Either Solaris supports atomic_sub and negating v is wrong, or atomic_sub needs to become atomic_add like the uint64_t operation.

matthewvon:
eleveldb.cc get_nif_string() is "safe" today because it is only called with a hard coded 1024 for max_size.  Therefore the danger of allocating that space on the stack is not too great.  But what if someone decides to reuse that for something big, like a stored/generated protocol buffer descriptor?  This could blow the stack (how big are Erlang stacks for that matter). 

Safe thing to do is either always "new" a buffer of max_size (slow), or create a fixed buffer of 1024 and use it if max_size is <= to that constant but create buffer via "new" if someone starts passing in bigger stuff.  Be sure to delete [] buffer if using created, but not if using stack.

Note, I am also assuming the function gets changed to pass a reference, not a pointer, to "out".  Safety!  Or at least check to make sure "out" is not NULL ... but reference is better.

c_src/workitems.cc
```diff
@@ -691,20 +874,20 @@ char* EncodeVarint64(char* dst, uint64_t v) {
  */
 work_result RangeScanTask::operator()()
```
matthewvon, line 875: Please see second catch block.  github will not let me comment on line 1063 of new code.  ErlUtil::formatBinary() can send a throw.  What happens to a throw that occurs during a catch?  I don't know.

matthewvon:
ErlUtil.cc checkTerm() might as well be checking that env is not null too.  If going to be paranoid, might as well be thoroughly paranoid.

matthewvon:
ErlUtil.cc isString.cc is allocating an unbounded string on the stack.  What if that field has a video file as a binary string?  See previous comment about small fixed buffer with logic to grab bigger buffer via "new".

matthewvon:
ErlUtil.cc isString.cc decides whether a binary is a string or not based upon "isprint()".  That function is locale specific ... and likely we are not setting locale.  Guessing even UTF is not safe.

I would settle for UTF and ignore unicode (for now).  Maybe post an Issue for unicode work.

matthewvon:
ErlUtil.cc getAtom():  another case of dynamic allocation on stack instead of heap.

matthewvon:
ErlUtil.cc getAtom() is tolower() going to work on anything but ascii? locale?  UTF?

matthewvon:
One more stack based allocation:  getString() ... maybe it is time to create a small class that is your buffer.  It contains 1024 bytes of storage, or allocates from heap if more needed.  Then its destructor does the right thing with heap allocated storage.  Class instantiated on stack to give self contained storage opportunity to be fast, Fast, FAST.

c_src/build_deps.sh
```diff
@@ -8,7 +8,7 @@ if [ `uname -s` = 'SunOS' -a "${POSIX_SHELL}" != "true" ]; then
 fi
 unset POSIX_SHELL # clear it so if we invoke other scripts, they run as ksh as well
 
-LEVELDB_VSN="end-to-end/timeseries"
+LEVELDB_VSN="tags/2.1.3"
```
matthewvon: Let's change this to "2.0.9".  That version has all the sequential data improvements from "develop" branch.  2.0 and 2.1 branches have same API at this time.

c_src/workitems.h
erikleitch, line 577: Indeed.  Done.

erikleitch:
get_nif_string() had all sorts of problems (why pass in a buffer that's
1024 and only fill it if the string is <= 256?  why not just resize the
passed string and use it as the buffer?)

It turns out this function was used only once, in a vestigial bit of code
that stores the global data directory, which is only used for the data
dictionary that I deleted entirely from this branch.

So I've just deleted get_nif_string(), the options parsing code that
exercised it, the corresponding GLOBAL_DATA_DIR atom, and the corresponding
member from EleveldbOptions

c_src/workitems.cc
erikleitch, line 875: The second catch block calls
ErlUtil::formatBinary(unsigned char\* buf, size_t size),  which does not
throw.  It just formats the passed buf to look like an erlang term, and
returns the formatted string.

(You're thinking of
ErlUtil::formatBinary(ErlNifEnv\* env, ERL_NIF_TERM term),
which can throw.  And a throw during a catch would  just throw another
exception that needs to be caught)

So this is not an unprotected throw.

erikleitch:
When I wrote ErlUtil, I wasn't sure at the time how it was going to be
used, so it's a bit needlessly versatile from an interface perspective.
You can use it three ways:

1) as a bunch of static function calls, which is mostly how we use it, i.e.:

std::string str = ErlUtil::getString(env, term);

2) as a class instantiated with an ErlNifEnv*, i.e.:

ErlUtil eu(env);
std::string str = eu.getString(term);

3) and as a class instantiated with both an ErlNifEnv\* and an ERL_NIF_TERM:

Erlutil eu(env, term);
std::string str = eu.getString();

That's why there are three different versions of most class methods.  The
no-arg versions call checkEnv() AND checkTerm() to make sure we have a
valid internal environment and term members.  The term-only versions call
only checkEnv() to make sure we have a valid internal environment member,
and the static versions just used the passed env and term arguments without
checking for any internal members.

So the idea is that we check the environment only when no environment is
provided,  we check the term only when no term is provided, and we check
both when neither is provided, but that functionality is separated out into
checkEnv(), which only checks the env, and checkTerm(), which only checks
the term.

c_src/build_deps.sh
erikleitch: Done

erikleitch:
Added a StringBuf struct to ErlUtil that has a minimum fixed buffer, and
allocates a heap buffer on resize if size > minsize.  This is now used in
isString(), getString() and getAtom().

erikleitch:
Added a StringBuf struct to ErlUtil that has a minimum fixed buffer, and
allocates a heap buffer on resize if size > minsize.  This is now used in
isString(), getString() and getAtom().

erikleitch:
Added a StringBuf struct to ErlUtil that has a minimum fixed buffer, and
allocates a heap buffer on resize if size > minsize.  This is now used in
isString(), getString() and getAtom().

erikleitch:
A good point.  The tolower option was only used (in theory) to allow users
to send atoms 'True' and 'False' and have these interpreted as booleans in
the C++ layer.

Since erlang only treats atoms 'true' and 'false' as booleans anyway, I've
just removed this code altogether (so that isBool() will return false for
atoms 'True' and 'False', and true for atoms 'true' and 'false')

erikleitch:
It's a good point.  I'm not quite sure what the 'right' thing to do here
is.  I've modified this to use xlocale + locale-aware isprint_l(), but the
question is really whether we need this code at all.  Allowing binaries to
be interpreted as strings was really just a convenience for me during
development (and maybe for formatting human-readable output when printing
error messages).

This functionality is not really needed anywhere in the production code
path, so one alternative is to leave it as it is until and when we need it
to behave in a particular way, or just remove it, and stick to erlang's
narrow definition of string as lists, binaries as binaries.

Opinions?

erikleitch:
Agreed that this is not portable behavior.  Automatically being able to
interpret erlang binaries as strings was really just a convenience during
code development (and for formatting human-readable messages), but this
behavior is not required for production code.

To avoid this ambiguity, I've removed any code from ErlUtil that allows
binaries to be interpreted as strings by default.  ErlUtil::isString() now
only returns true if the term is an erlang atom, or if enif_get_string()
would return 1.

Since we have asserted for TS that field names must be strings, and also
that they must be represented as erlang binaries, I've added
ErlUtil::getBinaryAsString() that does exactly that -- inspects an erlang
binary and returns it as a null-terminated string, without making any
assumptions about encoding or character sets.

c_src/StringBuf.cc
```diff
@@ -0,0 +1,110 @@
...
+{
+    // If size is <= bufSize_, do nothing
+
+    if(size > bufSize_) {
+        heapBuf_    = (char*)realloc(heapBuf_, size);
+        bufPtr_     = heapBuf_;
+        bufSize_    = size;
+    }
+
+    dataSize_ = size;
```
matthewvon: This is incorrect.  Look at constructor StringBuf(size).  There is no data, only resize to make sure buffer is sufficient.  Setting dataSize implies valid user data present.  No, this does not break any existing code.  But I like this class and suggest it could get wider use in the future.  Need to fix for future usage.

erikleitch: Well, it was a conscious choice actually, to cover external users accessing
the buffer directly and writing into it, without updating dataSize_.
(That's the mode it's currently used in for reading strings, for which the
dataSize_ is irrelevant and the null terminator is used as the end of data
marker.)

But I agree it's not well-defined behavior in general, so I'm happy to make
the other choice too, which is to assert dataSize_ == 0 until someone says
otherwise.

I've committed this change.

matthewvon:
+1

PR #165, opened 2015-10-23
Fixed detail.hpp so that it should compile under Solaris.  However, i…

erikleitch:
…t raises the issue of why we are calling these functions with uint types when the underlying atomic functions expect ints, and whether that implies any potential overflow


matthewvon:
+1

PR #168, opened 2015-11-05
Propagated recent SQL changes that apparently led to unadvertised cha…

erikleitch:
…nges to the elevledb interface ('integer' -> 'sint64', 'float' -> 'double', 'binary' -> 'varchar')

It was decided that fundamental datatypes (ints and floats) would be renamed at the SQL interface.  An unadvertised side effect of this change is that all the type names changed in the erlang-eleveldb interface too, which breaks all the C++ filtering code.  This PR includes support for the changes that were apparently made:

what used to be called 'binary' is now called 'varchar'
what used to be called 'float' is now called 'double'
what used to be called 'integer' is now called 'sint64'

I've modified the code to support the new naming conventions, and have modified the unit tests (sut.erl) to test the new naming conventions.

### DIFF
--- a/c_src/extractor.cc
+++ b/c_src/extractor.cc
@@ -75,12 +75,36 @@ DataType::Type Extractor::tsAtomToType(std::string tsAtom, bool throwIfInvalid)
 {
     DataType::Type type = DataType::UNKNOWN;
 
-    if(tsAtom == "binary") {
+    // Used to be 'binary', now it's 'varchar'
+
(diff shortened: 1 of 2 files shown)


matthewvon:
@erikleitch  +1 on code.  While I agree with Doug, I do not agree with his timing.  Please file an Issue against eleveldb that details which "old" names should be removed.  The cleanup should happen AFTER first ship.

PR #170, opened 2015-11-17
Removed support for old datatypes: integer, binary and float.  These …

erikleitch:
…were renamed sint64, varchar and double

I had left support in place in the C++ layer for the old typenames, just to protect against some erlang code still referencing the old names.  We decided to remove this to see what breaks, so this PR is quite simple:

1) three clauses have been removed from extractor.cc

2) all erlang unit tests in test/sut.erl that referenced the old datatypes have either been removed, or renamed to use the new datatypes


hazen:
:+1: 82ea4f1

erikleitch:
Also checked that riak_test ts_api test passes with the new branch

PR #171, opened 2015-12-03
Feature/bugfix filtercopy

erikleitch:
…reaming_start.  This avoids the issue of having to copy the erlang filter specification for later evaluation

**Context:**
When this code was originally developed, no type information was passed down from erlang about the fields used in the filtering.  This type information is required to build the expression tree that evaluates the filter in C++, which meant that the filter specification couldn't be parsed until the first data record was read (from which I could infer the datatype information).  In short, the original erlang filter spec was stored and parsed later.  

It was not, however, safely copied using enif_make_copy(), etc., and the code could therefore segfault if erlang garbage-collected the terms before the filter was parsed in the worker thread.

Since then, we have changed the erlang code to pass type information about the fields, so the filter can now be parsed during setup while the original erlang terms are still in scope.  This is the essence of this PR -- I've removed the delayed filter parsing from the worker thread in workitems.cc to the originating NIF call in eleveldb.cc.

Changes are:
1. added some notes to eleveldb.cc.  The code has not changed
2. extractor.cc:  The only substantive change is line 354, which sets expr_fields_[field] to whatever the filter specification says it is.  This is a map that originally contained the datatypes inferred from parsing the encoded data, which is later used in the data extraction (cf ExtractorMsgpack::extract() ) to compare against the datatypes used to build the filter.  Overwriting it here lets me use the rest of the extractor code unmodified.
3. workitems.cc: Two substantive changes:  
   - I've modified RangeScanTask to parse the filter in the constructor.  This constructor is called in eleveldb.cc:streaming_start while the relevant erlang terms are still in scope, thus eliminating the need to copy them.
   - I've removed the filter parsing from RangeScanTask::operator().  This was previously a needlessly complicated two-step process where I parsed the first data record to get the type information, then parsed the filter.  This code is no longer needed and has been removed
4. sut.erl: unit tests had to be changed because we've changed the behavior of the C++ code to throw errors on unsupported operators (e.g., '< ' used with binaries) rather than quietly return an empty key list.  This means that the corresponding null tests in sut.erl have to trap the error rather then check for an empty return list.

### DIFF
--- a/c_src/eleveldb.cc
+++ b/c_src/eleveldb.cc
@@ -551,11 +551,17 @@ ERL_NIF_TERM parse_streaming_option(ErlNifEnv* env, ERL_NIF_TERM item,
 
         } else if (option[0] == eleveldb::ATOM_RANGE_FILTER) {
 
-	    // Store the rangefilter spec and environment to parse later
+            //------------------------------------------------------------
+            // For the filter terms, opts is used merely as a
+            // convenience container.  The filter will be parsed in
+            // RangeScanTask constructor before these terms go out of
+            // scope of the NIF, so we do not need to safe copy them
+            // here
+            //------------------------------------------------------------
 
-	    opts.useRangeFilter_  = true;
-            opts.rangeFilterSpec_ = option[1];
+            opts.useRangeFilter_  = true;
             opts.env_             = env;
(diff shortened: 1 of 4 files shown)


paulplace:
+1 76f2f22

PR #174, opened 2015-12-14
Perf/test release candidate

erikleitch:
Added sync_put/sync_write functions to eleveldb to do the work of the leveldb write in the NIF call itself (rather than async_put, which expects the caller to block on an asynchronous reply, or put, which calls async_write and blocks in eleveldb for the reply).  

This is half of a last-minute performance optimization for TS1.0 (the other half is in riak_kv, where riak_kv_vnode:handle_command calls sync_put instead of async_put for the TS W1C path).

This could be improved by avoiding the object creation altogether and performing the leveldb write directly in the NIF code instead of using the WriteTask infrastructure, but I wanted to make as few untested deviations from async_put as possible, given the timeframe.

And it is understood that use of this function is very risky, since it provides no guarantees to the erlang scheduler about execution time.  The plan is to incorporate MvM's changes to check what leveldb is about to do before using a synchronous put, for TS1.1

### DIFF
--- a/c_src/eleveldb.cc
+++ b/c_src/eleveldb.cc
@@ -72,6 +72,7 @@ static ErlNifFunc nif_funcs[] =
 
     {"async_open", 3, eleveldb::async_open},
     {"async_write", 4, eleveldb::async_write},
+    {"sync_write", 4, eleveldb::sync_write},
     {"async_get", 4, eleveldb::async_get},
 
     {"async_iterator", 3, eleveldb::async_iterator},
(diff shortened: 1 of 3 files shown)


matthewvon:
I give a plus one to eleveldb.h and eleveldb.cc.

Do want to note that sync_put() could be written with stack objects instead "new" to improve performance even more.

erikleitch:
I've gone ahead and reworked sync_write to use stack objects, per Matthew's suggestion.

matthewvon:
Plus one to stack objects and C code in general.

PR #175, opened 2015-12-17
Fixing iterator memory leak in streaming folds code, also rare binary…

erikleitch:
… leak

This is a fix for a memory leak in the streaming folds code that emerged in stress testing the query path yesterday.  Leveldb iterator was being allocated but not freed in the range scan.

I've made two changes in this PR:

1) I've fixed the obvious iterator leak by wrapping it in auto_ptr.  This will delete the iterator when it goes out of scope, either on error or on exit

2) I've identified another possible rare memory leak, which could occur if no further keys matched after a batch was sent (because the enif_release_binary() call was keyed to the value of out_offset, which is reset each time a batch is sent).  This is now explicitly marked as allocated when enif_alloc_binary() is called, and released whenever allocated.

I have tested the TS1.0 release code with these changes, and summarize them below:

leaktests

Note that the original tests had only probed the iteration code (exercised whenever a range scan is executed against TS primary key conditions), and not the filtering code that I added for secondary key searches.  The summary plot above include tests with:
- no fix to the iterator leak (red line)
- leak fixed, only primary key queries (green line)
- leak fixed, primary + secondary queries (cyan line)

and for completeness, since none of this has previsouly been stress-tested, I've tested the put path:
- TS put only (blue line)

I see no evidence for significant residual leaks on the query path, and although the put path is much less stable (presumably because it is dominated by garbage collection of erlang terms), there is no obvious linear trend with time

### DIFF
--- a/c_src/workitems.cc
+++ b/c_src/workitems.cc
@@ -911,7 +911,7 @@ work_result RangeScanTask::operator()()
     read_options.fill_cache = options_.fill_cache;
     read_options.verify_checksums = options_.verify_checksums;
 
-    leveldb::Iterator* iter        = m_DbPtr->m_Db->NewIterator(read_options);
+    std::auto_ptr<leveldb::Iterator> iter(m_DbPtr->m_Db->NewIterator(read_options));
     const leveldb::Comparator* cmp = m_DbPtr->m_DbOptions->comparator;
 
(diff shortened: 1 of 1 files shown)


paulplace:
+1 e8bac8a

PR #177, opened 2016-01-05
Fixed potential memory leak on error in RangeScanTask::operator()

erikleitch:
catch clause on error didn't free allocated msg_env (or binary for returned values, for that matter).  This PR adds cleanup of these variables on exceptional return.


matthewvon:
+1  release of binary is redundant at this time since it is only set after this block.  But redundant safety is more future proof.  Leave it.

PR #178, opened 2016-01-25
Mv cache object warming

matthewvon:
Add cache_object_warming flag to eleveldb options.  Default is true.

Also add code to build_deps.sh to execute "git submodule update --init" if BASHO_EE environment variable is set to 1.  This enables non-open source code.


paulplace:
+1 56f6e10

PR #180, opened 2016-02-18
FIX: OTP release 18 does not match required regex R16|17

fogfish:
assembly of eleveldb is fails with OTP 18.0 or later

```
rm -Rf deps
make

ERROR: OTP release 18 does not match required regex R16|17 
```

The challenge is cuttlefish project that strictly enforces R16|17 release, the recent develop branch of that project relaxes the requirements.


henry-hz:
maybe point to the 2.0.5 tag instead of develop branch.

PR #183, opened 2016-03-04
Finish TS batching

macintux:
Other than accidental whitespace fixes, the only change here is to export `sync_write/4` for use by vnodes when batches of data arrive.

Required for basho/riak_kv#1366


macintux:
@borshop merge

PR #186, opened 2016-04-01
add fold() for options within eleveldb_repair().

matthewvon:
Original code for eleveldb_repair() lost a line for converting Erlang options into the eleveldb::Options structure.  The defaults for eleveldb::Options are sufficient for normal database / Riak vnodes, but not for Basho's leveldb tiered storage feature.  This branch corrects the issue.

More details:

https://github.com/basho/leveldb/wiki/mv-fix-repair-options

### DIFF
--- a/c_src/eleveldb.cc
+++ b/c_src/eleveldb.cc
@@ -1126,10 +1126,12 @@ eleveldb_repair(
     const ERL_NIF_TERM argv[])
 {
     char name[4096];
-    if (enif_get_string(env, argv[0], name, sizeof(name), ERL_NIF_LATIN1))
+    if (enif_get_string(env, argv[0], name, sizeof(name), ERL_NIF_LATIN1)
+        && enif_is_list(env, argv[1]))
     {
         // Parse out the options
         leveldb::Options opts;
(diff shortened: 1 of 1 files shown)


erikleitch:
@matthewvon This all seems sensible, and I've verified that:
- I can reproduce the problem as you describe
- your one-line fix to fold over the options fixes the problem

My only comment is this:

We know that default values for the Options structure can lead to incorrect repair and data loss.  What I don't have a feel for is what combinations of options in general can lead to this situation, and whether or not it's possible to detect that incompatible options have been specified.  I just don't know enough about how leveldb tiered storage works.    Guess I'll go read about that now...

If it were possible, then in addition to parsing the options in eleveldb:repair(), we would ideally check that the options as-specified can't lead to an error condition and report an error if they can.  

Thoughts on whether this even makes sense/ is possible?

paulplace:
+1

The code change looks good. I recommend adding your reproduction steps to the wiki.

erikleitch:
+1

I'm happy with the code changes.  @matthewvon and I have been talking about general solutions to mitigate this sort of error going forward, but that will take longer to implement (if it's even possible)

PR #187, opened 2016-04-05
Mv aae segfault

matthewvon:
eleveldb prefetch logic for iterator move operations failed to disable prefetch upon reaching the end of key space.  This lead to two symptoms, deadlock and bad return values, that sometimes resulted in a segfault.  

Detail discussion found here:  https://github.com/basho/leveldb/wiki/mv-aae-segfault


paulplace:
+1 2b1de14

PR #188, opened 2016-04-05
Modify spec for eleveldb:fold to permit first_key option.

jonmeredith:
Required for upcoming hashtree PR in riak_core.

`eleveldb:fold` also accepts a {first_key, <<binary>>} option that controls call to `iterator_move(Itr, Seek)`.


JeetKunDoug:
:+1: c4c50ef

PR #190, opened 2016-04-15
Don't use deprecated erlang:now/0 in tests

lemenkov:
Subj.

Signed-off-by: Peter Lemenkov lemenkov@gmail.com


bsparrow435:
Thank you!

PR #191, opened 2016-04-15
Correct race condition against iterator management flag [JIRA: RIAK-2528]

matthewvon:
This branch corrects a race condition that allows a foreground thread to overwrite an essential flag set by a background thread.

Details here: https://github.com/basho/leveldb/wiki/mv-aae-iterator-hang


paulplace:
+1 732464f

bashopatricia:
create jira issue

Basho-JIRA:
Reviewed and merged per Doug 

**_[posted via JIRA by Patricia Brewer]**_

PR #192, opened 2016-04-22
fix lager for eleveldb

hmmr:
- rebar.config: mention lager_transform, depend on lager;
- eleveldb.erl: drop unused import;
- eleveldb.erl: pass received ErrMsg via "~p".


erikleitch:
+1 c981b5d

Thanks Andrei  -- been meaning to do this myself for a while!

PR #195, opened 2016-05-06
Type conversion of bool to ??? too tricky with Solaris CAS.

matthewvon:
A recent branch added atomic compare and set operations for the bool variable m_PrefetchStarted.  This did not compile on Solaris.  Solaris does have an 8-bit version of its compare set.  But I did not find a document that guaranteed bool would be 8-bits.  I could still use the 8-bit CAS, but would then have to assume all platforms were little endian.

Decided to take the safe and fast path.  Changed m_PrefetchStarted to uint32_t.  Then matched all the compare_and_swap() call parameters to match the existing compare_and_swap(volatile uint32_t *, int, int) function.

All platforms compile and test just fine now.

### DIFF
--- a/c_src/eleveldb.cc
+++ b/c_src/eleveldb.cc
@@ -775,7 +775,8 @@ async_iterator_move(
     const ERL_NIF_TERM& action_or_target = argv[2];
     ERL_NIF_TERM ret_term;
 
-    bool submit_new_request(true), prefetch_state;
+    bool submit_new_request(true);
+    int prefetch_state;      // not bool for Solaris CAS
 
     ReferencePtr<ItrObject> itr_ptr;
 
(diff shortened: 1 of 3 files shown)


c_src/eleveldb.cc
```diff
@@ -860,10 +861,11 @@ async_iterator_move(
         // using compare_and_swap has a hardware locking "set only if still in same state as before"
         //  (this is an absolute must since worker thread could change to false if
         //   hits end of key space and its execution overlaps this block's execution)
+	int cas_temp((eleveldb::MoveTask::PREFETCH_STOP != action )  // needed for Solaris CAS
```
paulplace: Please fix the indentation on these lines. The inserted lines (864, 865, and 868) use tab characters, but the surrounding code uses spaces.

paulplace:
+1 d989ced

PR #196, opened 2016-06-09
Mv iter debug1

matthewvon:
This branch is only for debug purposes on a production system.  The code will get tagged and built, but later removed.  It will NOT merge to "develop".

The key goal is to track the execution of a long running iterator.  If not running, log its forced removal at the time of node shutdown.


paulplace:
I examined the code in commit c6b9963, built it, and ran unit tests on OS X. All looks good.

matthewvon:
Code included in temporary eleveldb 2.0.20 release.  This code will NOT merge to develop branch.

matthewvon:
Just in case, keeping this code in production for a while.

PR #200, opened 2016-06-21
Cache Valid() state of iterator

matthewvon:
Isolate leveldb iterator Valid() state. This allows the iterator to be destroyed on background thread, releasing leveldb snapshot. The foreground thread is then able to read the Valid() state at its leisure.


paulplace:
+1 05719a3

paulplace:
+1 a7c191c

PR #203, opened 2016-07-05
integrate open source expiry features into eleveldb option processing

matthewvon:
Automatically add an ExpiryModuleOS object to leveldb::Options if any one of the 3 expiry options seen.  Only need one instance of the object and it will auto destruct after last pointer released.

### DIFF
--- a/c_src/build_deps.sh
+++ b/c_src/build_deps.sh
@@ -8,7 +8,7 @@ if [ `uname -s` = 'SunOS' -a "${POSIX_SHELL}" != "true" ]; then
 fi
 unset POSIX_SHELL # clear it so if we invoke other scripts, they run as ksh as well
 
-LEVELDB_VSN="mv-expiry"
+LEVELDB_VSN="expiry_beta1"
 
 SNAPPY_VSN="1.0.4"
 
--- a/c_src/build_deps.sh
+++ b/c_src/build_deps.sh
@@ -8,7 +8,7 @@ if [ `uname -s` = 'SunOS' -a "${POSIX_SHELL}" != "true" ]; then
 fi
 unset POSIX_SHELL # clear it so if we invoke other scripts, they run as ksh as well
 
-LEVELDB_VSN="expiry_beta1"
+LEVELDB_VSN=""
 
 SNAPPY_VSN="1.0.4"
 
--- a/c_src/build_deps.sh
+++ b/c_src/build_deps.sh
@@ -8,7 +8,7 @@ if [ `uname -s` = 'SunOS' -a "${POSIX_SHELL}" != "true" ]; then
 fi
 unset POSIX_SHELL # clear it so if we invoke other scripts, they run as ksh as well
 
-LEVELDB_VSN=""
+LEVELDB_VSN="mv-expiry"
 
 SNAPPY_VSN="1.0.4"
 
--- a/c_src/eleveldb.cc
+++ b/c_src/eleveldb.cc
@@ -43,6 +43,7 @@
 #include "leveldb/perf_count.h"
 #define LEVELDB_PLATFORM_POSIX
 #include "util/hot_threads.h"
+#include "leveldb_os/expiry_os.h"
 
 #ifndef INCL_WORKITEMS_H
     #include "workitems.h"
@@ -133,6 +134,9 @@ ERL_NIF_TERM ATOM_TIERED_SLOW_LEVEL;
 ERL_NIF_TERM ATOM_TIERED_FAST_PREFIX;
 ERL_NIF_TERM ATOM_TIERED_SLOW_PREFIX;
 ERL_NIF_TERM ATOM_CACHE_OBJECT_WARMING;
+ERL_NIF_TERM ATOM_EXPIRY_ENABLED;
+ERL_NIF_TERM ATOM_EXPIRY_MINUTES;
+ERL_NIF_TERM ATOM_WHOLE_FILE_EXPIRY;
 }   // namespace eleveldb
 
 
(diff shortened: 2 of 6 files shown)


c_src/build_deps.sh
```diff
@@ -8,7 +8,7 @@ if [ `uname -s` = 'SunOS' -a "${POSIX_SHELL}" != "true" ]; then
 fi
 unset POSIX_SHELL # clear it so if we invoke other scripts, they run as ksh as well
 
-LEVELDB_VSN=""
+LEVELDB_VSN="mv-expiry"
```
paulplace: Is this still needed? I think the mv-expiry branch of leveldb has been merged to develop.

matthewvon: Good catch ... that is one of the classic eleveldb check-in screw ups.

Restored to default ""

c_src/eleveldb.cc
```diff
@@ -459,6 +465,45 @@ ERL_NIF_TERM parse_open_option(ErlNifEnv* env, ERL_NIF_TERM item, leveldb::Optio
...
+            else
+            {
+                if (NULL!=opts.expiry_module.get())
+                    ((leveldb::ExpiryModuleOS *)opts.expiry_module.get())->expiry_enabled = false;
+            }   // else
+        }   // else if
+        else if (option[0] == eleveldb::ATOM_EXPIRY_MINUTES)
+        {
+            unsigned long minutes(0);
+            if (enif_get_ulong(env, option[1], &minutes))
```
paulplace, line 485: Should we log (syslog?) an error if the enif_Xxx function fails?

c_src/eleveldb.cc
```diff
@@ -459,6 +465,45 @@ ERL_NIF_TERM parse_open_option(ErlNifEnv* env, ERL_NIF_TERM item, leveldb::Optio
                 opts.cache_object_warming = false;
         }
 
+        else if (option[0] == eleveldb::ATOM_EXPIRY_ENABLED)
+        {
+            if (option[1] == eleveldb::ATOM_TRUE)
+            {
+                if (NULL==opts.expiry_module.get())
```
paulplace, line 472: Should checking/creating the expiry_module member be moved outside the ATOM_TRUE conditional? That way, we set the expiry_enabled member, regardless of whether the expiry_module member was previously created. Then we don't have to worry about the order in which the various expiry options are declared or what their defaults are in the leveldb::ExpiryModuleOS class.

The same question/comment applies to the other expiry options below.

c_src/eleveldb.cc
```diff
@@ -568,8 +613,13 @@ async_open(
     eleveldb_priv_data& priv = *static_cast<eleveldb_priv_data *>(enif_priv_data(env));
 
     leveldb::Options *opts = new leveldb::Options;
+    opts->expiry_module.assign(priv.m_ExpiryModule.get());
+
     fold(env, argv[2], parse_open_option, *opts);
+
     opts->fadvise_willneed = priv.m_Opts.m_FadviseWillNeed;
+    if (NULL==priv.m_ExpiryModule.get() && NULL!=opts->expiry_module.get())
```
paulplace, line 621: I am unclear on the purpose of this code. A few lines above, you assigned opts->expiry_module from priv.m_ExpiryModule, so they should be the same. If there's some reason why they may not be the same, a comment here would be helpful.

priv/eleveldb.schema
```diff
@@ -239,3 +239,31 @@
...
+  {datatype, flag},
+  hidden
+]}.
+
+%% @doc Minutes until object expires.  Gives number of minutes
+%% a stored key/value will stay within the database before automatic
+%% deletion.  Zero disables expiry by age (minutes) feature
+{mapping, "leveldb.expiry_minutes", "eleveldb.expiry_minutes", [
+  {default, 0},
+  {datatype, integer},
```
paulplace, line 257: When you parse the expiry_minutes option in eleveldb.cc, you use the enif_get_ulong() function. Is that compatible with declaring it as an "integer" here? I'm not sure of the type system in this schema, which is why I'm asking (e.g., perhaps integers are always long and unsigned).

c_src/eleveldb.cc
matthewvon, line 485: Created issue to track suggestion for future. This is of same nature as your suggestion to have pthread creation errors post to syslog.  I think I will do all of these at once.

https://github.com/basho/eleveldb/issues/204

c_src/eleveldb.cc
matthewvon, line 472: No.  Keep in mind that we do not know what order options will arrive.  And that cuttlefish often passes default values to common options.

I am intentionally not creating the object unless absolutely necessary.  It is highly likely that the default options will get passed ... and defaults currently leave expiry disabled.  So only creating the object if non-defaults seen.  However, I do update the object if a different non-default option caused its creation.

c_src/eleveldb.cc
matthewvon, line 621: Line 616:  has a previous vnode opened and created an expiry module?  Give me the previously created object if it exists.
Line 621:  if there was no previously created expiry module and one suddenly appears due to this fold, make it the default.

This squirrel code defends against the future possibility of having different expiry modules by vnode and yet still a global default.  Not likely ... but possible given that the future code is not yet written.

priv/eleveldb.schema
matthewvon, line 257: cuttlefish is its own little world.  I do copy/paste of previous options as my base.  The enif_get_XxX() functions used in eleveldb.cc are defined as filtering out of range values.  Therefore my defense is at that layer, not here or other weirdness in eleveldb.erl.

paulplace:
+1 2f58433

PR #207, opened 2016-07-15
Disable cuttlefish dependency except for buildbot tests

matthewvon:
eleveldb does not care about cuttlefish dependencies.  However, there currently exist three active branches of eleveldb due solely to other Riak code that is specific about cuttlefish.

The branch comments out the rebar.config "deps" block.  And it comments out the two lines that enable cuttlefish unit tests within test/eleveldb_schema_tests.erl.  The comments use a tag that is removed via a sed command during buildbot unit tests.  Then the default cuttlefish on develop is used to validate the eleveldb .schema files.


paulplace:
+1 8bb5e52

PR #208, opened 2016-07-20
add lz4 as option to eleveldb option block [JIRA: RIAK-2707]

matthewvon:
This updates the eleveldb.schema file to accept "snappy" and "lz4" as independent options for the leveldb.compression variable.  The default value remains "snappy", but end users now have the option to enable lz4 compression.


matthewvon:
Note:  in leveldb, the compression setting in LOG file will be 
0 - disabled
1 - Snappy
2 - LZ4
(anything else is a severe bug!!)

c_src/eleveldb.cc
```diff
@@ -353,14 +357,15 @@ ERL_NIF_TERM parse_open_option(ErlNifEnv* env, ERL_NIF_TERM item, leveldb::Optio
         }
         else if (option[0] == eleveldb::ATOM_COMPRESSION)
         {
-            if (option[1] == eleveldb::ATOM_TRUE)
-            {
+            if (option[1] == eleveldb::ATOM_ON || option[1] == eleveldb::ATOM_TRUE
+                || option[1] == eleveldb::ATOM_SNAPPY )
                 opts.compression = leveldb::kSnappyCompression;
-            }
```
fadushin, line 359: I ALWAYS use curly braces around block statements, even if they are one-liners.  That future proofs you from stupidity, and it doesn't make the code run any slower.  'twas the rule at Iona, and I always live by it.

matthewvon, line 359: My opinion on that same topic fluctuates.  Some days I agree with you and Iona.  Other days not.  No particular reason.  This was a case of that parent function getting too large and fewer lines seemed to make it seem clearer.  But I have heard (and even repeated) the argument that leaving the braces prepares the code for maintenance.

I will adjust if you feel strongly.

priv/eleveldb_multi.schema
```diff
@@ -116,8 +116,8 @@
 {mapping,
   "multi_backend.$name.leveldb.compression",
   "riak_kv.multi_backend", [
-  {default, on},
-  {datatype, flag},
+  {default, snappy},
+  {datatype, [{atom, on}, {atom, off}, {atom, snappy}, {atom, lz4}]},
```
fadushin: There is a minor b/w compatibility issue here we should think about it.  The previous type was flag, so 'true' and 'false' are allowable, and maybe even others ('0', and '1'?   Not sure, I am not a cuttlefish expert).  So if someone used 'true' before, then they will have a start failure on an upgrade.

priv/eleveldb.schema
```diff
@@ -158,8 +158,8 @@
 %% changed at any time, but will not impact data on disk until
 %% next time a file requires compaction.
 {mapping, "leveldb.compression", "eleveldb.compression", [
-  {default, on},
-  {datatype, flag},
+  {default, snappy},
+  {datatype, {enum, [on, off, true, false, snappy, lz4]}},
```
fadushin: Woops.  Same comment as below.

JeetKunDoug:
After a quick discussion with @angrycub we'd rather see this as something like:

```
eleveldb.compression = on | off 
eleveldb.compression.algorithm = snappy | lz4
```

Cuttlefish convention generally favors "simple" settings used in combination vs. single complex settings like what is proposed here. Additionally, we then end up with 2 Erlang settings (is compression on or off, and which algorithm should we use) as independent variables, rather than conflating the two into a single thing with on/off/true/false/snappy/lz4 options.

priv/eleveldb.schema
matthewvon: Actually not the same.  I failed to copy this line to multibackend.  The above line appears to work with all old values ... cuz I tested true and false.  Original cuttlefish expected on/off then generated true/false.

priv/eleveldb_multi.schema
matthewvon: Failed to update this from eleveldb.schema.

matthewvon:
somebody know how to code that in cuttlefish?  'cuz the result is the same by the time it hits eleveldb (I believe from other options).

c_src/eleveldb.cc
JeetKunDoug, line 359: Personally, I do honestly - the whole "oops, that statement was supposed to be part of the "if/true" block but instead ended up just executing because I forgot a curly brace" bugs are nasty (see http://embeddedgurus.com/barr-code/2014/03/apples-gotofail-ssl-security-bug-was-easily-preventable/).

src/eleveldb.erl
```diff
@@ -297,7 +297,7 @@ option_types(open) ->
      {block_size_steps, integer},
      {paranoid_checks, bool},
      {verify_compactions, bool},
-     {compression, bool},
+     {compression, any},  %% how to do a list of atoms?
```
fadushin: Looks like there is a `validate_type` function in this module, which we could augment to take a list/enumeration of valid atoms and then check to make sure that the value is in the enumeration.  Something like:

```
validate_type({_Key, [snappy, lz4]}, snappy) -> true;
validate_type({_Key, [snappy, lz4]}, lz4) -> true;
```

JeetKunDoug:
@matthewvon I can help you with the change to add `algorithm` as a thing separate from the compression. I'm sure there are other examples of the above floating around we can use.

c_src/eleveldb.cc
matthewvon, line 359: Done and committed.

priv/eleveldb_multi.schema
matthewvon: See new macros using "translate".

src/eleveldb.erl
matthewvon: I am still clueless.  Need erlang help here.

bashopatricia:
create jira issue

priv/eleveldb.schema
```diff
@@ -163,6 +163,25 @@
...
+
+{translation,
+ "eleveldb.compression",
+ fun(Conf) ->
+  Setting = cuttlefish:conf_get("leveldb.compression", Conf),
+    case Setting of
+      false -> false;
+      true ->
+        cuttlefish:conf_get("leveldb.compression.algorithm", Conf, snappy);
+      _Default -> snappy
```
JeetKunDoug: This clause will never get hit and should be removed - leveldb.compression has a default value of `on` and can only be true or false by the time this fun is called. Not sure why it's also in the Bitcask translation we used as an example, but ¯\_(ツ)_/¯

matthewvon: Line removed.

test/eleveldb_schema_tests.erl
```diff
@@ -102,6 +102,57 @@ override_schema_test() ->
...
+
+    cuttlefish_unit:assert_config(Config4, "eleveldb.compression", snappy),
+
+
+    ok.
+
+
+
+
 multi_backend_test() ->
```
JeetKunDoug, line 154: Although it looks like it _should_ work correctly, can we add tests for the multi-backend as well for compression settings?

matthewvon, line 154: Done.  Pushed.

JeetKunDoug:
+1 cc0be0a

Basho-JIRA:
[~matthewv] Should this story for LZ4 compression for Riak move to "Needs Review" for someone to pick it up and review it? Didn't want to just move it forward without confirming with you.  

**_[posted via JIRA by Patricia Brewer]**_

Basho-JIRA:
LZ4 options within riak.conf have passed review.  Doug requests a new Jira ticket for updating Riak dependency lists with a new eleveldb tag.  That tag will be ready tomorrow morning (Wed 8/9). 

**_[posted via JIRA by Matthew Von-Maszewski]**_

PR #210, opened 2016-07-27
update expiry cuttlefish params to match notes from cv

matthewvon:


erikleitch:
+1, though I'll leave it for @angrycub to close.

This seems like a simple change of three parameter names, from:
- leveldb.expiry_enabled to leveldb.expiration
- leveldb.expiry_minutes to leveldb.expiration.retention_time, with a change of units to {datatype, {duration, m}}
- leveldb.whole_file_expiry to leveldb.expiration.mode, with a change of units to {datatype, {flag, whole_file, normal}}

Name changes make sense to me, appropriate tests have been added for default values in single and multi-backend uses, as well as overriding default parameter vals.

angrycub:
:+1: +1 Looks good to me.

PR #211, opened 2016-08-02
Merge expiry from develop branch into TS (ready)

erikleitch:
<hr>

**Overview**

This branch incorporates handling of options in eleveldb to support the new cluster-wide leveldb expiry mechanism.  This includes cuttlefish changes to eleveldb.schema as well as C++ code to support parsing those options in eleveldb and instantiating the expiry module in leveldb.

<hr>

**Changes include:**

**c_src/eleveldb.cc** 
- code has been changed to include leveldb_os/expiry_os.h header file
- new atoms have been added for expiry_enabled, expiry_minutes and whole_file_expiry options
- code has been added to parse_open_option, to parse the corresponding expiry options.
  
  NB: Per notes in the code, I have changed the logic that instantiates the leveldb::ExpiryModuleOS object to create it if _any_ expiry option is specified.   This differs from the original version in mv-expiry-schema2 branch, which only instantiated the object if non-default values were specified.  I believe that this is either an oversight, or an accident waiting to happen.  If the leveldb ExpiryModuleOS constructor were to change the default values of certain parameters for example, then it would cause those options to be lost if they were specified prior to enabling expiry

**priv/eleveldb.schema**
- Schema now contains mappings for expiry parameters eleveldb.expiry_enabled, eleveldb.expiry_minutes and eleveldb.whole_file_expiry
- Per @matthewvon 's suggestion, I have changed the default eleveldb.compression from 'snappy' to 'lz4'

**c_src/refobjects.cc** 
- A log comment was added in LevelIteratorWrapper::LogIterator() when debugging a Meltwater ticket (iterator hang, which we now believe has been fixed).  This was commented out in the develop branch, but the uncommented version had been incorporated into riak_ts-develop.  I have removed the line entirely to avoid confusion.

**build_deps.sh**
- LEVELDB_VSN has been changed from 2.0.22 to the leveldb tag containing the expiry code (this is still a WIP, but should be 2.0.24 when done)

**rebar.config**
- Some comments were added regarding buildbot unit tests on the develop branch.  I have left those in

<hr>

**Tests**
- All unit tests pass against this branch
- I've also tested this branch in riak_test for basic TS put and query functionality (including simple range-scan and secondary filtering), and it behaves as expected
- I've also run a series of put/get KV  latency tests against this branch and found that latencies agree with the TS1.3 baseline


c_src/build_deps.sh
```diff
@@ -8,7 +8,11 @@ if [ `uname -s` = 'SunOS' -a "${POSIX_SHELL}" != "true" ]; then
 fi
 unset POSIX_SHELL # clear it so if we invoke other scripts, they run as ksh as well
 
+<<<<<<< HEAD
 LEVELDB_VSN="2.0.22"
+=======
+LEVELDB_VSN=""
```
matthewvon: This going to change to 2.0.24 now that the tag is finally available?

c_src/eleveldb.cc
```diff
@@ -245,6 +253,8 @@ class eleveldb_priv_data
     EleveldbOptions m_Opts;
     leveldb::HotThreadPool thread_pool;
     leveldb::HotThreadPool stream_thread_pool;
+    leveldb::ExpiryPtr_t m_ExpiryModule;    // only populated if expiry options seen
```
matthewvon: m_ExpiryModule is not present in final code version.  Dead code ... simply erase this line and next.

c_src/eleveldb.cc
```diff
@@ -500,6 +517,45 @@ ERL_NIF_TERM parse_open_option(ErlNifEnv* env, ERL_NIF_TERM item, leveldb::Optio
                 opts.cache_object_warming = false;
         }
 
+        else if (option[0] == eleveldb::ATOM_EXPIRY_ENABLED)
+        {
+            if (option[1] == eleveldb::ATOM_TRUE)
+            {
+                if (NULL==opts.expiry_module.get())
+                    opts.expiry_module.assign(new leveldb::ExpiryModuleOS);
+                ((leveldb::ExpiryModuleOS *)opts.expiry_module.get())->expiry_enabled = true;
```
matthewvon: This section of code does not look like what you sent previously.  Is diff wrong, or am I looking at wrong code?

c_src/eleveldb.cc
```diff
@@ -1280,7 +1336,7 @@ streaming_start(ErlNifEnv * env,
         work_item = new RangeScanTask(env, reply_ref, db_ptr.get(),
                                  start_key, end_key_ptr, opts, sync_handle->sync_obj_);
     } catch(std::runtime_error& err) {
```
matthewvon, line 1338: Nightmare.  Hope this is coded correctly or Erlang crashes.  There is a reason for no try/catch blocks.

c_src/refobjects.cc
```diff
@@ -416,11 +416,6 @@ LevelIteratorWrapper::LogIterator()
     struct tm created;
 
     localtime_r(&m_IteratorCreated, &created);
```
matthewvon: might as well remove this too

matthewvon:
As before, I have severe issues with try/catch blocks.  Not going to press the issue until TS and KV merge.

I will take responsibility for moving the adjusted expiry option code in eleveldb.cc to mainline.

+1

PR #213, opened 2016-08-20
Mv mem fences

matthewvon:
This is one of two parallel branches named mv-mem-fences.  The second branch is within the leveldb repository.  The two branches must be used together.  

The branches address an extremely rare race condition that causes a protective assert() to fire within eleveldb's refobjects.cc.

A detailed description of the bug and fix is found here:

https://github.com/basho/leveldb/wiki/mv-mem-fences

### DIFF
--- a/c_src/build_deps.sh
+++ b/c_src/build_deps.sh
@@ -8,7 +8,7 @@ if [ `uname -s` = 'SunOS' -a "${POSIX_SHELL}" != "true" ]; then
 fi
 unset POSIX_SHELL # clear it so if we invoke other scripts, they run as ksh as well
 
-LEVELDB_VSN="mv-mem-fences"
+LEVELDB_VSN=""
 
 SNAPPY_VSN="1.0.4"
 
--- a/c_src/build_deps.sh
+++ b/c_src/build_deps.sh
@@ -8,7 +8,7 @@ if [ `uname -s` = 'SunOS' -a "${POSIX_SHELL}" != "true" ]; then
 fi
 unset POSIX_SHELL # clear it so if we invoke other scripts, they run as ksh as well
 
-LEVELDB_VSN=""
+LEVELDB_VSN="mv-mem-fences"
 
 SNAPPY_VSN="1.0.4"
 
--- a/c_src/eleveldb.cc
+++ b/c_src/eleveldb.cc
@@ -193,7 +193,7 @@ struct EleveldbOptions
     bool m_FadviseWillNeed;
 
     EleveldbOptions()
-        : m_EleveldbThreads(71),
+        : m_EleveldbThreads(7),
           m_LeveldbImmThreads(0), m_LeveldbBGWriteThreads(0),
           m_LeveldbOverlapThreads(0), m_LeveldbGroomingThreads(0),
           m_TotalMemPercent(0), m_TotalMem(0),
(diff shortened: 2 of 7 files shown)


c_src/eleveldb.cc
```diff
@@ -274,9 +274,18 @@ ERL_NIF_TERM parse_init_option(ErlNifEnv* env, ERL_NIF_TERM item, EleveldbOption
         else if (option[0] == eleveldb::ATOM_LIMITED_DEVELOPER_MEM)
         {
             if (option[1] == eleveldb::ATOM_TRUE)
+            {
                 opts.m_LimitedDeveloper = true;
+
+                // lower thread count too if developer stuff.
+                //  each thread has 2Mbytes to 8Mbytes of stack
+                if (71==opts.m_EleveldbThreads)
```
paulplace, line 282: Do we have this 71 value defined as a macro somewhere? Would be nice to use a manifest constant rather than a hard-coded magic number. Also, do we want to always drop it to 7 if it's greater than 7? That is, should the conditional be "if (7 > opts.m_EleveldbThreads)"? Perhaps not if we want to respect a purposely-set value.

matthewvon, line 282: some day I will #define or const int the value.  but not today.  and no, if not 71 then we respect that someone is playing with values and get out of the way

c_src/eleveldb.cc
```diff
@@ -705,6 +714,7 @@ async_write(
     ERL_NIF_TERM result = fold(env, argv[2], write_batch_item, *batch);
     if(eleveldb::ATOM_OK != result)
     {
+        delete batch;
```
paulplace, line 719: Would be good to add a comment (perhaps above where the batch variable is defined) indicating that the lifetime of the batch variable is managed by the WriteTask, created below. We only need to delete it here since we haven't yet handed it to the WriteTask object.

Even better (or at least more modern C++) would be to make batch an auto_ptr and call auto_ptr<>.release() in the WriteTask constructor. That way we make clear who owns the WriteBatch object at any given moment.

paulplace:
I don't think the RefObject class in refobjects.h is needed any longer.

UPDATE: I just noticed that the ctor/dtor of RefObject update perf counters, so please ignore the previous comment.

c_src/build_deps.sh
```diff
@@ -8,7 +8,7 @@ if [ `uname -s` = 'SunOS' -a "${POSIX_SHELL}" != "true" ]; then
 fi
 unset POSIX_SHELL # clear it so if we invoke other scripts, they run as ksh as well
 
-LEVELDB_VSN=""
+LEVELDB_VSN="mv-mem-fences"
```
paulplace: This will need to be set back to "" when you merge.

c_src/refobjects.h
```diff
@@ -73,16 +73,15 @@ class ErlRefObject : public RefObject
...
 
-    // 1 once InitiateCloseRequest starts,
-    // 2 other pointers to "this" released
-    // 3 final RefDec and destructor executing
-    volatile uint32_t m_CloseRequested;
-
     leveldb::port::Mutex   m_CloseMutex;        //!< for condition wait
     leveldb::port::CondVar m_CloseCond;         //!< for notification of user's finish
 
 protected:
```
paulplace: You introduced the GetCloseRequested() method to ensure that m_CloseRequested is read cleanly, so I recommend making m_CloseRequested private and making the few direct references to m_CloseRequested (in DbObject) use calls to GetClosedRequested() instead.

c_src/eleveldb.cc
matthewvon, line 719: done

c_src/refobjects.h
matthewvon: Done.  Thought I had already removed all read references.  Proves the power of "private:".

paulplace:
+1 64b17b1

PR #218, opened 2016-09-13
Make multi_backend compression settings commented

nickelization:
These should only be appearing in riak.conf for documentation, as they
won't be needed unless a user is setting up multi_backend. (Even then
they will probably want to change the backend name to something other
than "name".)


fadushin:
+1 14cff4f

PR #221, opened 2016-10-17
Switch to reference counted pointers instead of naked pointers.

matthewvon:
Switch the few remaining naked pointer uses to reference counted pointers.  This fills some remaining race condition holes (of which one recently seen).

Also, move LevelIteratorWrapper from being a reference counted object.  It is now directly embedded within ItrObject.  This is most likely the source of a recent SegFault.  The ItrObject::Shutdown function released the LevelIteratorWrapper reference pointer which set up a race condition.

Finally, old iterator logging code that was commented out was removed.


fadushin:
Preliminary thumbsup, pending some system level testing we are undertaking currently.

+1

fadushin:
Reviewed latest changes since last +1 in person with MVM
+1

PR #222, opened 2016-10-18
Dynamic decision to swap threads or not on Erlang Write operation

matthewvon:
eleveldb swaps each Erlang request from an Erlang scheduler thread to a local eleveldb worker thread.  This is to prevent Erlang scheduler collapses.  Write operations do not actually need to swap all the time.  This branch creates the capability of dynamic selection to swap threads or not.

Detailed discussion is here:

https://github.com/basho/leveldb/wiki/mv-nonblocking-write2


c_src/build_deps.sh
```diff
@@ -8,7 +8,7 @@ if [ `uname -s` = 'SunOS' -a "${POSIX_SHELL}" != "true" ]; then
 fi
 unset POSIX_SHELL # clear it so if we invoke other scripts, they run as ksh as well
 
-LEVELDB_VSN=""
+LEVELDB_VSN="mv-nonblocking-write"
```
paulplace, line 11: Standard reminder to set this back to "" before merging.

c_src/eleveldb.cc
```diff
@@ -703,40 +717,67 @@ async_write(
...
-    }   // if
+        leveldb::WriteOptions* opts = new leveldb::WriteOptions;
+        fold(env, argv[3], parse_write_option, *opts);
+
+        eleveldb::WorkTask* work_item = new eleveldb::WriteTask(env, caller_ref,
+                                                                db_ptr.get(), batch, opts);
+        eleveldb_priv_data& data = *static_cast<eleveldb_priv_data *>(enif_priv_data(env));
+        if(false == data.thread_pool.Submit(work_item))
+        {
+            // work_item contains "batch" and the delete below gets both memory allocations
```
paulplace, line 772: work_item also takes ownership of the opts variable that was allocated on the heap; it's probably worth updating the comment so that the next person in this code doesn't have to hunt down that fact

matthewvon:
this branch was replaced by mv-nonblocking-writes3 ... a manual rebase

PR #223, opened 2016-10-20
Replace msgpack encoding with ttb

erikleitch:
## **Overview**

This is one of two related PRs to replace msgpack with ttb as the default on-disk encoding format for TS.  The other is https://github.com/basho/riak_kv/pull/1510.

Msgpack was originally chosen as the encoding format because in early TS development (before the query pipeline was even written), micro-benchmarking of decoding riak objects in C++ for secondary filtering found that msgpack was significantly faster than naive implementation of the ei decoding library for ttb. In the context of the end-to-end query pipeline, however, that decoding in C++ represents less than 10% of the total query latency. Far and away the largest contribution to the latency is converting encoded records back to erlang terms in riak_kv_qry:decode_results() when they are returned to riak from eleveldb. In this context, the binary_to_term BIF is significantly more efficient than our implementation of msgpack decoding -- up to 30% faster for queries of 100K rows, and increasing with the size of the query.

You can find more discussion of the original micro-benchmarks, the query pipeline, and performance testing of this branch at https://github.com/basho/internal_wiki/wiki/Performance-Improvements-for-Riak#ts1.5.

The purpose of these PRs is to improve query performance by replacing msgpack with ttb encoding as the default on-disk format for TS data.
## **Changes in this repo**

**Summary** 
The class `RangeScanTask` defined in `workitems.h/cc`, is responsible for the streaming fold that executes a TS query.  When a secondary filter is passed in with the range scan options, the `RangeScanTask` constructor parses it into an expression tree that is vetted against each record on disk by decoding the record, populating the tree with values for the fields specified in the filter, and evaluating the tree.  Decoding the record is accomplished by an `Extractor` object which understands how to unpack a riak object and decode the data payload (defined in `extractor.h/cc`).

In previous versions of TS, only msgpack encoded records were legal, and the inherited class `ExtractorMsgpack` was used to decode all records.  That extractor lives inside the `RangeScanTask` object and was created when `RangeScanTask` was instantiated.  `ExtractorMsgpack` uses the class `CmpUtil` for all msgpack-related decoding operations, via the cmp library.

This PR adds a new inherited extractor, `ExtractorErlang` that decodes ttb-encoded records.  It uses the class `EiUtil` for all ttb-related decoding operations, via the ei library.   `EiUtil` is a direct analog of `CmpUtil` and is structured in much the same way.  

In this PR, `RangeScanTask` has been modified no longer to instantiate a single extractor object, but a _map_ of extractors for all valid encoding types.  When records are decoded, the encoding byte recorded with the riak object is used as an index into the map, to obtain the correct extractor for the current record.  In this way, the on-disk encoding no longer needs to be specified by the caller, and any mix of valid encodings can be processed seamlessly by the fold.

**Details**
`c_src/DataType.h` TTB encoding uses a special type ("small big") for large-integer encoding that must be handled separately from existing types, and a corresponding enum has been added here.

`c_src/EiUtil.h/cc` As mentioned in the summary above, `EiUtil` collects together all operations needed for decoding and formatting TTB-encoded data.  It is directly analogous to `CmpUtil` for msgpack, and is structured accordingly.  It is based on the erlang ei binary decoding interface, but one of the main methods used in the fold, `EiUtil::skipTerm` has been written from scratch, to avoid having to use ei decode calls to advance the index pointer when skipping over fields that are not part of the filter.  This replacement is responsible for a 4x speedup, making `ExtractorErlang` as fast as `ExtractorMsgpack` (see <a href=#perf>Performance</a> below)

`c_src/eleveldb.cc` Because we no longer make assumptions about the on-disk encoding format, but inspect each record as it's encountered, we no longer need the caller to specify an encoding.  The unused encoding atoms have therefore been removed, and the option -- if specified -- is simply ignored.

`c_src/ErlUtil.cc` When constructing error messages, I've changed the default behavior of `ErlUtil::formatTerm` to attempt to format binaries as human-readable strings, since most 'binaries' used in TS are actually strings to begin with.

`c_src/StringBuf.cc` A method has been added to return a C++ string version of the internal char\* buffer.

`c_src/exceptionutils.h` Not used in production code, but always useful during development, I'm adding a simple macro that allows dumping messages to disk.

`c_src/extractor.cc` The new inherited extractor `ExtractorErlang` has been added to this file, along with the `ExtractorMap` class for managing a map of extractors.  Base-class `riakObjectCanBeParsed` method has been modified to decode the encoding byte and check it against known valid encodings, and `extractRiakObject` has been moved into the baseclass.  In both `ExtractorErlang` and `ExtractorMsgpack` an additional check has been added that stops the decoding when all fields referenced by the filter have been encountered.  This further optimizes performance for long records by only decoding the part of the object we need to evaluate the filter.

`c_src/filter.h` I realized that by allowing NULLs, we had introduced a logical error in evaluation of binary AND and OR expressions.  When NULLs were not allowed, we checked if each operand to the binary expressions had a value, and the expression evaluated to false if either didn't.  Since has_value() is now used to indicate a valid NULL value, these checks must be removed, and both operands must be evaluated.

`c_src/filter_parser.h/cc` Because the single extractor has been changed to an `ExtractorMap` type, every method of filter_parser has been changed to take an `ExtractorMap&` argument instead.  When fields are added by a filter condition, they are now added to every extractor in the map, since we do not know up-front which ones will be needed during the fold.

`c_src/workitems.h` As discussed in the summary above, `RangeScanTask` has been modified to manage a map of extractors, rather than a single extractor, and `RangeScanTask::DoWork` now looks up the right extractor in the map for each disk record on the fly.  All code related to parsing the deprecated encoding type from the filter options has been removed.

`test/sut.erl` Changes to test filtering over interleaved msgpack and ttb-encoded records, and to provide coverage for decoding any integer data type that may be encountered.  Previous version of sut.erl tested integer operations, but only for small ints.  For both msgpack and ttb, integers can be encoded as a variety of types depending on value, and the current test therefore explicitly tests all regimes to exercise this behavior.  Finally, tests for invalid encoding specification have been removed, since specifying an encoding is no longer required, and an invalid encoding therefore no longer generates an error.
## **Compatibility**

**Backwards Compatibility** Because the new code uses the on-disk encoding flag to determine which extractor to use, data previously written in msgpack format will continue to be correctly interpreted in TS1.5.  As noted in https://github.com/basho/riak_kv/pull/1510, the riak_kv code is also agnostic as to the encoding, and will correctly handle records returned to erlang in either format.

**Downgrade Compatibility** Because the riak_kv code is agnostic (already handles TTB data for KV records), older versions of TS will correctly interpret TS data written with ttb encoding.  As noted in https://github.com/basho/riak_kv/pull/1510, however, if a customer downgrades to TS1.4 after writting data with TS1.5, they would need this version of eleveldb or the ttb-encoded records would be skipped.  Older versions of riak_kv will work with this version of eleveldb, since the deprecated encoding option is quietly ignored if specified.
## <a name=perf>**Performance**</a>

In addition to the functional testing coverage in `sut.erl`, extensive testing of this branch is detailed at https://github.com/basho/internal_wiki/wiki/Performance-Improvements-for-Riak#ts1.5.  I reproduce a summary plot here:

msg_ttb_filter_cmp

The plot shows latencies for TS1.4 (left panel), this branch (middle panel), and the difference in ms (right panel), for a wide range of query sizes using secondary filtering.  The plot demonstrates two important points:
- on the short-query side (Nrow < 100), where the latency is dominated by the actual eleveldb fold operation, the difference is consistent with zero, demonstrating that the ttb decoding is as efficient as the msgpack decoding in C++
- on the large-query side (Nrow > 100), where the latency is dominated by decoding of returned records in erlang, this branch is more than 30% faster than TS1.4 -- for example the reduction in latency for 100K-row queries is 700 ms (out of ~2 seconds).

### DIFF
--- a/c_src/EiUtil.cc
+++ b/c_src/EiUtil.cc
@@ -1,0 +1,1141 @@
+#include "DataType.h"
+#include "EiUtil.h"
+
+#include "exceptionutils.h"
+
+#include <sstream>
+#include <cmath>
+#include <iomanip>
+
+#include <climits>
+
+using namespace std;
+
+using namespace eleveldb;
+
+#define THROW_ERR ThrowRuntimeError("Not a valid EI term")
+
+#define FN_DEF(retType, name, body)             \
+    retType EiUtil::name()                      \
+    {                                           \
+        checkBuf();                             \
+        return name(buf_, &index_);             \
+    }                                           \
+                                                \
+    retType EiUtil::name(char* buf, int* index) \
+    {                                           \
+        body;                                   \
+    }
+    
+//=======================================================================
+// Initialize static maps of conversion functions here
+//=======================================================================
+
+std::map<DataType::Type, EI_CONV_UINT8_FN(*)>  
+EiUtil::uint8ConvMap_ = EiUtil::constructUint8Map();
+
+std::map<DataType::Type, EI_CONV_INT64_FN(*)>  
+EiUtil::int64ConvMap_ = EiUtil::constructInt64Map();
+
+std::map<DataType::Type, EI_CONV_UINT64_FN(*)> 
+EiUtil::uint64ConvMap_ = EiUtil::constructUint64Map();
+
+std::map<DataType::Type, EI_CONV_DOUBLE_FN(*)> 
+EiUtil::doubleConvMap_ = EiUtil::constructDoubleMap();
+
+//=======================================================================
+// A macro for declaring a template convert specialization
+//=======================================================================
+
+#define CONVERT_DECL(typeTo, typeFrom, getName, validation)             \
+    namespace eleveldb {                                                \
+        template<>                                                      \
+        typeTo EiUtil::convert<typeTo, typeFrom>(char* buf, int* index) \
+        {                                                               \
+            typeFrom val;                                               \
+            try {                                                       \
+                val = getName(buf, index);                              \
+                validation;                                             \
+            } catch(...) {                                              \
+                ThrowRuntimeError("Object of type " << typeOf(getType(buf, index), buf, index) << " typestr = " << EiUtil::typeStrOf(buf, index) \
+                                  << " can't be represented as a " << #typeTo << " using " << #getName << " typeFrom = " << #typeFrom); \
+            }                                                           \
+            return (typeTo) val;                                        \
+        }                                                               \
+    }                                                                   \
+                                                                        \
+//=======================================================================
+// A macro for constructing a map of conversion functions to the
+// specified type
+//=======================================================================
+
+#define CONSTRUCT_EI_CONV_MAP(typeTo)                                   \
+    /* int types */                                                     \
+    convMap[DataType::INT64]  = EiUtil::convert<typeTo, int64_t>; \
+                                                                        \
+    /* float types */                                                   \
+    convMap[DataType::DOUBLE] = EiUtil::convert<typeTo, double>; \
+                                                                        \
+    /* Boolean types */                                                 \
+    convMap[DataType::BOOL]   = EiUtil::convert<typeTo, bool>;   
+                                                                        
+EiUtil::EiUtil()
+{
+    initialize(0, 0);
+}
+
+EiUtil::EiUtil(char* buf)
+{
+    initialize(buf, 0);
+}
+
+EiUtil::~EiUtil() {};
+
+void EiUtil::initialize(char* buf, int index)
+{
+    buf_     = buf;
+    index_   = index;
+    version_ = -1;
+
+    version_ = getVersion();
+}
+
+FN_DEF(int, getVersion,
+
+       int version = -1;
+       
+       if(ei_decode_version(buf, index, &version))
+           THROW_ERR;
+       
+       return version;
+    )
+
+FN_DEF(int, getTupleHeader,
+
+       int arity = 0;
+       
+       if(ei_decode_tuple_header(buf, index, &arity))
+           THROW_ERR;
+       
+       return arity;
+    )
+
+FN_DEF(int, getListHeader,
+
+       int arity = 0;
+       
+       if(ei_decode_list_header(buf, index, &arity))
+           THROW_ERR;
+       
+       return arity;
+    )
+
+FN_DEF(ei_term, decodeTerm,
+
+       ei_term term;
+       if(ei_decode_ei_term(buf, index, &term) < 0)
+           THROW_ERR;
+       
+       return term;
+    )
+
+FN_DEF(std::string, printTerm,
+
+       char* s=0;
+       
+       if(ei_s_print_term(&s, buf, index))
+           THROW_ERR;
+       
+       std::string ret(s);
+       
+       if(s)
+           free(s);
+       
+       return ret;
+    )
+
+/**.......................................................................
+ * Return the type of this object
+ */
+FN_DEF(int, getType,
+       int size=0;
+       int type=0;
+        
+       if(ei_get_type(buf, index, &type, &size) < 0)
+           THROW_ERR;
+        
+       return type;
+    )
+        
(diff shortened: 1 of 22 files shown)


c_src/EiUtil.cc
```diff
@@ -0,0 +1,1689 @@
...
+FN_DEF(std::string, formatString,
+
+       std::ostringstream os;
+       std::string str = getString(buf, index);
+       os << "\"" << str << "\"";
+       os << " (aka [";
+       unsigned int size=str.size();
+       for(unsigned int i=0; i < size; i++) {
+           os << (int)str[i];
+           if(i < size-1)
```
paegun: not a must, but since operations are optimized for comparing to zero and especially because there is no need to evaluate size and no need to subtract one per loop, this is cheaper expressed as (moved up a line):

``` c
  if (i > 0) { os << ", "; }
```

c_src/EiUtil.h
```diff
@@ -0,0 +1,212 @@
...
+#define ELEVELDB_EIUTIL_H
+
+/**
+ * @file EiUtil.h
+ * 
+ * Tagged: Wed Sep  9 17:32:28 PDT 2015
+ * 
+ * @version: $Revision: $, $Date: $
+ * 
+ * @author /bin/bash: username: command not found
```
paegun: Whatever template tool is being used, should be using `whoami` or `echo $USER` as `username` isn't present by default on OS X or Ubuntu. In any case, should be the author's name :)

Similar for version, but from scanning other files w/i c_src, this is consistent.

c_src/ErlUtil.cc
```diff
@@ -1175,6 +1188,23 @@ std::string ErlUtil::formatBinary(unsigned char* buf, size_t size)
...
+std::string ErlUtil::formatAsString(char* buf, size_t size)
+{
+    return formatAsString((unsigned char*) buf, size);
+}
+
+std::string ErlUtil::formatAsString(unsigned char* buf, size_t size)
+{
+    std::ostringstream os;
+
+    for(unsigned iByte=0; iByte < size; iByte++)
```
paegun, line 1206: Since this loop effectively casts each unsigned char to char, this whole function could instead be written as:

``` c
    return std:string((char*)buf);
```

paegun:
Only found 2 places where the code may be changed for:
1. a minor micro optimization, prepending "," in generating a CSV instead of appending.
2. a code simplification and clearer case of casting unsigned char to char while converting a `unsigned char*` to `std::string` .

+1 bdf9d08

c_src/EiUtil.cc
erikleitch: Done  (Indeed, I made little effort to optimize the formatXXX methods, since they are only used when formatting error messages, so don't impact performance in normal use)

I've now changed all the formatXXX methods that print lists in both EiUtil and ErlUtil to behave the same way

c_src/EiUtil.h
erikleitch: Good catch -- yeah, these are auto-generated by some emacs macros I use to write C++ code.  They also put some outdated CVS-style tags in the header.  I've removed these from all the class files that had them, and replaced with a brief summary of the purpose of each class, creation date where available, and author email.

c_src/ErlUtil.cc
erikleitch: This one I'm going to leave as-is.  The important point is that ErlUtil::formatAsString() is called on erlang-formatted binaries that are not in general null-terminated (Things like <<"f2">>, which erlang will format as 2 bytes).  So passing the char buf directly to the string constructor can result in undefined behavior (garbage characters on the ends of otherwise-valid strings).  That's why this method copies the bytes into the ostringstream and explicitly null-terminates it before passing it to the string constructor.

paegun:
+1 dfa757b

PR #226, opened 2016-11-10
Port submit_to_thread_queue() from riak_ts-develop to develop

hazen:
This refactor of message queuing from the TS looks like a nice optimization.  I have found many merge conflicts related to the two code bases differing, so I'm proposing we simply agree to use this function on both sides of the fence.


paegun:
+1 a350fcf

A call to `priv.thread_pool.Submit(move_item)` remains w/i `async_iterator_move` due to the move_item intentionally not being deleted as the work_item is deleted w/i the refactor.

hazen:
Yes, that case has a different return value which is probably why it was not converted

PR #227, opened 2016-11-10
RTS-1448: Merge 2.0.32 with TS

hazen:
Hopefully this is the last merge of KV `riak-2.2` over to the TS side

### DIFF
--- a/BASHO_RELEASES
+++ b/BASHO_RELEASES
@@ -1,6 +1,15 @@
-github.com tag 2.0.30 - October 11, 2016
+github.com tag 2.0.31 - November 1, 2016
 ----------------------------------------
  - version shipped with Riak 2.2
+ - includes leveldb 2.0.31 (mv-no-md-expiry & mv-tuning8)
+mv-ref-hardening:   - series of thread hardening changes
+                      related to AAE use of iterators. Biggest
+                      fix was isolating AAE using one thread to
+                      close while another was still moving, then
+                      defending against it
+
+github.com tag 2.0.30 - October 11, 2016
+----------------------------------------
  - includes leveldb 2.0.30 (mv-delayed-bloom)
 
 github.com tag 2.0.28 - September 7, 2016
--- a/BASHO_RELEASES
+++ b/BASHO_RELEASES
@@ -1,6 +1,23 @@
-github.com tag 2.0.30 - October 11, 2016
+github.com tag 2.0.32 - November 8, 2016
+----------------------------------------
+** additional race condition hardening when faced with two threads on same iterator
+**  (one iterating async_iterator_move() and one closing async_iterator_close())
+ - wrap async_iterator_move operations with locked CloseMutex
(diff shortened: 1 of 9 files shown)


erikleitch:
This PR merges changes made in KV eleveldb branch since the last KV/TS eleveldb merge, pretty much all of which are to do with @matthewvon's efforts to debug AAE hangups using leveldb iterators.  

According to Matthew, a lot of the eleveldb code pertaining to reference counting (and thread synchronization in general) was modified in the process -- not necessarily for cause, but for debugging -- and those changes were simply never reverted once a workaround for the problem was identified (adding a mutex lock around async_iterator_move).   Note that the workaround was tested and shipped with the unreverted changes in place.

**My summary of key changes in this PR:**

  * Some of the changes are merely syntactic: all XXXTask constructors have been modified to use the `ReferencePtr(const ReferencePtr &rhs)` constructor rather than the `ReferencePtr(TargetT *_t)` constructor, for example.  I don't know why this particular change was made (see my comment above), but I also don't see that it does any harm.  

    _This particular change to the RangeScanTask constructor is the only one that directly touches any of the TS-specific code, but it does not change the behavior of that object._

  * `async_iterator_close` ReferencePtr assignment was moved inside `ItrObject::RetrieveItrObject()`, as part of  mitigating the async_iterator_move/close race condition by pushing the assignment for both functions into a protected section where the close check is also performed.  Changes were made on this commit: https://github.com/basho/eleveldb/commit/71458ee08caf211c838009044172ecc9d8d0caa9.    A wrapper object ItrObjErlang has also been introduced in this branch to provide a spin lock for protecting access to the iterator pointer in RetrieveItrObject called from different threads.

    _I don't see any problems with this._

  * ItrObject::m_Iter has been changed from a `ReferencePtr<LevelDbWrapper>` object to `ItrObject::m_Wrap` a bare `LevelDbWrapper`, and most of the other deltas in this PR are a syntactic knock-on effect of doing this.  According to Matthew, there was not a specific motivator for this change, but it was also part of the effort to debug AAE hanging during iteration.  

    _This last change concerns me a bit, because I don't immediately see why this object was _originally_ made a reference-counted pointer, so I can't easily convince myself that we haven't changed the functionality by removing it.  But I am prepared to take it on faith that the reference-counted version of this object is no longer needed, per Matthew's claim, since these changes were also reviewed and tested for KV.  In the context of TS, changes to normal iterators should not impact any TS operations, since we don't use them._

hazen:
Thanks, @erikleitch.  Is there additional testing, perhaps, you would like to do to see if the change to the iterators have any unforeseen consequences?

erikleitch:
+1

I don't think so, in the sense that without knowing how exactly the ref counting was previously used, I can't off-the-cuff suggest a test that could provoke a failure around it.  

+1ing because I think the most pragmatic approach will be to rely on making sure that all riak_test tests pass (presumably we know that they do on the KV branches) in the merged version.

hazen:
Thanks again, @erikleitch!

PR #229, opened 2016-11-21
Backport submit_to_thread_queue() to KV

hazen:
This refactor of message queuing from the TS side looks like a nice optimization. I have found many merge conflicts related to the two code bases differing, so I'm proposing we simply agree to use this function on both sides of the fence.


matthewvon:
+1 for merge once TS 1.5 ships

PR #231, opened 2016-11-24
RTS-1549: Merge 2.0.33 onto TS

hazen:
https://bashoeng.atlassian.net/browse/RTS-1549
Includes update of leveldb to 2.0.33
https://bashoeng.atlassian.net/browse/RTS-1544


erikleitch:
+1 

I'm happy with the (superficial) changes merged in from KV branch.  

The only substantial changes are a couple instances of returning a non-empty work_result in ItrCloseTask where previously there was no message.  This shouldn't affect erlang code upstream that is not currently expecting any response and is therefore not pattern-matching on one.

hazen:
Thanks @erikleitch!

PR #234, opened 2017-01-03
use https to pull deps

korry8911:
@javajolt

### DIFF
--- a/c_src/build_deps.sh
+++ b/c_src/build_deps.sh
@@ -32,7 +32,7 @@ MAKE=${MAKE:-make}
 
 get_dep_leveldb () {
     if [ ! -d leveldb ]; then
-        git clone git://github.com/basho/leveldb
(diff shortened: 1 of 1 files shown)


c_src/build_deps.sh
```diff
@@ -32,7 +32,7 @@ MAKE=${MAKE:-make}
 
 get_dep_leveldb () {
     if [ ! -d leveldb ]; then
-        git clone git://github.com/basho/leveldb
+        git clone https://github.com/basho/leveldb
```
matthewvon, line 35: Just curious, why the change.

korry8911: we can probably just close this... i was attempting to get all riak deps to use https instead of git since there are some problems when trying to use git... the project is a little more complicated than I thought and I don't think this is needed for now. Closing @matthewvon

PR #235, opened 2017-01-10
Mv bucket expiry2

matthewvon:


src/eleveldb_sup.erl
```diff
@@ -0,0 +1,36 @@
...
+%% -------------------------------------------------------------------
+-module(eleveldb_sup).
+
+-behaviour(supervisor).
+
+-export([start_link/0]).
+-export([init/1]).
+
+start_link() ->
+    supervisor:start_link({local, ?MODULE}, eleveldb_sup, []).
```
tburghart: This name should be added to the `registered` list in the .app.src file.

src/eleveldb_app.erl
```diff
@@ -0,0 +1,37 @@
...
+
+-behavior(application).
+
+-export([start/2, stop/1]).
+
+start(_Type, _Args) ->
+    case eleveldb_sup:start_link() of
+        {ok, Pid} ->
+            {ok, Pid};
+        Other ->
```
tburghart: Either return the result of `eleveldb_sup:start_link` directly, which is standard practice, or handle the possible returns from `supervisor:start_link`.
As written, `{error, {error, ...}}` is a possible result, which is undesirable for upstream reporting.

src/eleveldb.erl
```diff
@@ -335,7 +338,13 @@ validate_options(Type, Opts) ->
                             validate_type(KType, V)
                     end, Opts).
 
+-spec property_cache(string(), string()) -> ok.
+property_cache(_BucketKey, _Properties) ->
+    erlang:nif_error({error, not_loaded}).
 
+-spec set_metadata_pid(pid()) -> ok.
+set_metadata_pid(_Pid) ->
+    erlang:nif_error({error, not_loaded}).
```
tburghart: I would prefer a generic interface here to register an external service pid so we don't need a new function each time we do this sort of thing, and I don't expect this to be the only service we're going to want to call from eleveldb (think lager).

For example:

```erlang
-spec register_service(Type :: atom(), Pid :: pid(), Info :: term()) -> {ok, reference()} | {error, term()}.
-spec unregister_service(Type :: atom(), Key :: pid() | reference()) -> ok | {error, term()}.
```

In this example,
```erlang
    set_metadata_pid(P)
```
might be called as
```erlang
    register_service(eleveldb_metadata, P, [])
```

I'm not sure about the `unregister` side, just throwing something out for the case where we detect some service is unavailable. It seems a ref could be associated with a registration that could subsequently be used to unregister that service.

src/eleveldb.app.src
```diff
@@ -22,5 +22,6 @@
          %%  "older" compression by default
          {compression, snappy}
 
-        ]}
+        ]},
+   {mod, {eleveldb_app, []}}
  ]}.
```
tburghart: Add named servers started by the application to the `registered` list.

c_src/eleveldb.cc
```diff
@@ -73,13 +78,17 @@ static ErlNifFunc nif_funcs[] =
     {"async_iterator", 3, eleveldb::async_iterator},
     {"async_iterator", 4, eleveldb::async_iterator},
 
-    {"async_iterator_move", 3, eleveldb::async_iterator_move}
+    {"async_iterator_move", 3, eleveldb::async_iterator_move},
+
+    {"property_cache", 2, eleveldb::property_cache},
+    {"set_metadata_pid", 1, eleveldb::set_metadata_pid}
```
tburghart: Same as my comment in erleveldb.erl, I'd like this to be a more generalized mechanism.

src/eleveldb_metadata.erl
```diff
@@ -0,0 +1,110 @@
...
+            %% No recursion here; `system_continue/3' will be invoked
+            %% if appropriate
+            sys:handle_system_msg(Request, From, Parent,
+                                  ?MODULE, Debug, Fun);
+
+        callback_shutdown=Msg ->
+            sys:handle_debug(Debug, fun write_debug/3,
+                             ?MODULE,
+                             {in, Msg}),
+            ok;
```
tburghart: Should this not call `erlang:exit(shutdown)`?

src/eleveldb_metadata.erl
```diff
@@ -0,0 +1,110 @@
...
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+-module(eleveldb_metadata).
+
```
tburghart, line 23: I realize this is designed to handle raw messages from a NIF, but a protocol interface with lots and lots of doc comments seems like a good thing to have.

src/eleveldb_sup.erl
```diff
@@ -0,0 +1,36 @@
...
+
+init(_Args) ->
+    SupFlags = {one_for_one, 2, 5},
+    ChildSpecs = [{metadata, {eleveldb_metadata, start_link, []},
+                   permanent, brutal_kill, worker, [eleveldb_metadata]}],
```
tburghart: Why `brutal_kill`?  The process should be written to handle exit messages cleanly.

src/eleveldb_metadata.erl
```diff
@@ -0,0 +1,110 @@
...
+-export([callback_router/3]).
+
+%% Application messages are sent to this process directly from
+%% NIF-land instead of other Erlang modules, so this does not
+%% implement an OTP behavior.
+%%
+%% See http://erlang.org/doc/design_principles/spec_proc.html for
+%% information on these system message handlers.
+-export([system_continue/3, system_terminate/4,
```
tburghart: If this were a `gen_whatever` shutdown messages could unregister the service, as I've suggested elsewhere.

tburghart (dismissed):
I haven't gone into the C++ side at all, as I thing significant changes are needed on the Erlang side first.

So, this creates a circular dependency on `riak_core` (see xref output), which is a Very Bad Thing™, suggesting (strongly) that the model is wrong and that the metadata service should be running in `riak_core`, not in `eleveldb`.
Obviously, `eleveldb` needs to support registration of the service, but it doesn't need to run the service itself.

Needs unit tests, LOTS of them!  (https://github.com/basho/eleveldb/pull/240 might make that easier, were it to be merged)

At a minimum, and I'm not saying this is the right thing to do, a clean build requires:
* Dialyzer apps list `Makefile`.
* XRef exclusions in `rebar.config`.

tburghart:
@JeetKunDoug is there an associated PR in `riak_core`?  I see what I assume is your working branch, a review-only PR might be good.

JeetKunDoug:
Yea - was waiting to hear from @matthewvon about the viability of it before I PRed the riak_core change, but can WIP PR for review now...

JeetKunDoug:
riak_core changes are in basho/riak_core#898

tburghart:
Grrr, GitHub makes it look so simple ... do _NOT_ let it resolve the conflicts for you!

The `mv-bucket-expiry2-try-merge` branch has what you get if you do, and appears to be pulling in more than expected.

macintux:
There will need to be changes to reflect the `riak_core` code that's likely to be merged soon.

At the very least, the callback functions that handle the PID registration and receive the bucket properties must now return `ok`.

https://github.com/basho/riak_core/pull/898 for more details, especially the commentary at the top of https://github.com/basho/riak_core/blob/dr-th/service-poc/src/riak_core_info_service.erl

macintux:
* https://github.com/basho/riak_core/pull/898 <- merged
* https://github.com/basho/riak_core/pull/904 <- pending, adds `bucket_props` argument to the pid-handling callbacks

erikleitch:
(I thought @tburghart was reviewing this one, but the riak team said this morning that I should look at it too, so I'm adding a few notes here, despite not being sure whether this is ready for review, etc)

---
* **eleveldb_metadata.erl** Is there a deep reason why
    eleveldb:property_cache() is separately called from
    eleveldb_metadata.erl (instead of just having riak_core call
    `eleveldb:property_cache` directly)?  (Or is it leftover from an
    earlier architecture that has now changed?)

* In **router.cc**, is there a reason not to return an error atom (or
    message) if any of `add_metadata_pid`, `remove_metadata_pid`, or
    `property_cache` is called incorrectly? (currently, the code
    returns ATOM_OK for a number of these, whether or not there's an
    error).

  (Looks like these are in process of changing, so maybe this comment
  will become irrelevant)

* **router.cc** uses a global service callback object
  (`gBucketPropCallback`) to maintain the erlang pid to contact for
  bucket props.  This is used both from leveldb to request bucket prop
  info (via `leveldb_callback`) and from erlang to add/remove the pid
  (via `set_metadata_pid/remove_metadata_pid`), so simultaneous access
  from multiple thread contexts is possible.

  Shouldn't this resource be mutex-protected to ensure that the pid
  isn't accessed after `remove_metadata_pid` has been called, or that
  `ServiceCallback::GetPid` returns true after the pid has been set?

macintux:
Good questions Erik.

* I think it makes sense to have a translation function between the riak_core callback data and what the NIF expects. The former was designed to be general-purpose.
* `eleveldb_metadata` in general is a relic of an earlier architecture, so we could move that translation function into `eleveldb` proper, although for the moment I'd like to leave it so we don't have to yet again touch the riak_core PR.
* The NIFs need to return `ok` if, and only if, riak_core should continue servicing eleveldb metadata queries. The riak_core info service code shuts down the service process if any other return value is received. So, if there is an error that means that leveldb can't use the info service, then return an error.

macintux:
Ideally I think we'd allow eleveldb to report two categories of errors: permanent and transient. Future enhancement to the info service, perhaps.

matthewvon:
Added locking around ServiceCallback accessors per code review.

erikleitch (approved):

erikleitch (commented):
Oops - I think the spec for `eleveldb:property_cache/2` is now wrong (`eleveldb::property_cache` changed to return ok, or {error, any()}, but spec still reads: `-spec property_cache(string(), string()) -> ok.`)

matthewvon:
Wrong, maybe. In flux, absolutely.  Waiting on API lock down.

matthewvon:
This was a horrible merge.  Done by hand directly to develop.

PR #239, opened 2017-01-28
Test cleanup

tburghart:
This started with fixing tests to _NOT_ leave test files all over the place, but it quickly became apparent that many of the tests weren't running at all.

_Turn off whitespace diffs by appending `?w=1` (or `&w=1`, as appropriate) to the URL._

With these changes:

* All tests actually run, and pass.
* A few tests have been added that seemed appropriate.
* All tests write only in `/tmp` and clean up after themselves.
* All relevant cases of failures that would throw a `badmatch` are now enclosed in appropriate `?assert...` macros.
* The fundamental behavior of the tests hasn't been changed.

_Note: This does nothing about the cuttlefish kludge, sorry MvM._

### DIFF
--- a/src/eleveldb.erl
+++ b/src/eleveldb.erl
@@ -1,8 +1,6 @@
 %% -------------------------------------------------------------------
 %%
-%%  eleveldb: Erlang Wrapper for LevelDB (http://code.google.com/p/leveldb/)
-%%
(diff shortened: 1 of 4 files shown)


tburghart:
Closed and branch deleted because at least some builders, somewhere, still have issues with slashes in branch names.

PR #240 is the relevant replacement.

PR #240, opened 2017-01-28
Test cleanup

tburghart:
This started with fixing tests to _NOT_ leave test files all over the place, but it quickly became apparent that many of the tests weren't running at all.

_Turn off whitespace diffs by appending `?w=1` (or `&w=1`, as appropriate) to the URL._

With these changes:

* All tests actually run, and pass.
* A few tests have been added that seemed appropriate.
* All tests write only in `/tmp` and clean up after themselves.
* All relevant cases of failures that would throw a `badmatch` are now enclosed in appropriate `?assert...` macros.
* The fundamental behavior of the tests hasn't been changed.

_Note: This does nothing about the cuttlefish kludge, sorry MvM._

This PR replaces #239 with a branch name that _should_ be compatible with all builders.

### DIFF
--- a/src/eleveldb.erl
+++ b/src/eleveldb.erl
@@ -1,8 +1,6 @@
 %% -------------------------------------------------------------------
 %%
-%%  eleveldb: Erlang Wrapper for LevelDB (http://code.google.com/p/leveldb/)
-%%
-%% Copyright (c) 2010-2012 Basho Technologies, Inc. All Rights Reserved.
+%% Copyright (c) 2010-2017 Basho Technologies, Inc.
 %%
 %% This file is provided to you under the Apache License,
 %% Version 2.0 (the "License"); you may not use this file
@@ -19,6 +17,8 @@
 %% under the License.
 %%
 %% -------------------------------------------------------------------
+
+%% @doc Erlang NIF wrapper for LevelDB
 -module(eleveldb).
 
 -export([open/2,
@@ -49,11 +49,17 @@
 -on_load(init/0).
 
 -ifdef(TEST).
--compile(export_all).
+-export([
+    assert_close/1,
+    assert_open/1,
+    assert_open/2,
+    create_test_dir/0,
+    delete_test_dir/1,
+    terminal_format/2
+]).
 -ifdef(EQC).
 -include_lib("eqc/include/eqc.hrl").
--define(QC_OUT(P),
-        eqc:on_output(fun(Str, Args) -> io:format(user, Str, Args) end, P)).
+-define(QC_OUT(P), eqc:on_output(fun terminal_format/2, P)).
 -endif.
 -include_lib("eunit/include/eunit.hrl").
 -endif.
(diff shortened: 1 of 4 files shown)


matthewvon:
I have executed the buildbot automation against this branch.  There were 4 failing platforms out of 16.  This is the typical count.  I manually reviewed the failing four.  Two executed fine upon manual retest.  The other two (smartos1.8 and freebsd10) were typical resource problems.  Not going to sweat it.  I +1 from the automation viewpoint.

src/eleveldb.erl
```diff
@@ -395,105 +411,280 @@ validate_type(_, _)                                          -> false.
...
-    os:cmd("rm -rf /tmp/eleveldb.close_fold.test"),
-    {ok, Ref} = open("/tmp/eleveldb.close_fold.test", [{create_if_missing, true}]),
-    ok = eleveldb:put(Ref, <<"k">>,<<"v">>,[]),
-    ?assertException(throw, {iterator_closed, ok}, % ok is returned by close as the acc
-                     eleveldb:fold(Ref, fun(_,_A) -> eleveldb:close(Ref) end, undefined, [])).
+%% ===================================================================
+%% Exported Test Helpers
+%% ===================================================================
+
+-spec assert_close(DbRef :: db_ref()) -> ok | no_return().
```
JeetKunDoug, line 426: The "spec, comments, function" pattern you're using for some reason really messes with me - we almost always have doc comments above spec and function declarations. Can you please swap these around and, where they are public functions (maybe not too many) make them `@doc` comments?

JeetKunDoug: Only other thing (and again, general style note) is that left-aligned comments generally have 2 `%` characters, so all the comments above functions should be double-`%` - some sort of emacs-mode standard, and things can get messed up if you aren't consistent about it.

JeetKunDoug: All 42 tests pass locally... if you could do the minor cleanup stuff I'll +1.

JeetKunDoug (commented):
Just some stylistic/consistency notes on formatting - otherwise looks good.

src/eleveldb.erl
tburghart: I use a single `%` prefix when it's not published in edoc.
If it doesn't have a `%% @doc` comment, it's because it's not exported in all cases, and edoc will blow up on it.  Basically, anything within the `-ifdef(TEST).` block with any edoc comments causes failure (edoc's not very smart sometimes).

tburghart: The "spec, comments, function" pattern is not at all uncommon in Erlang code, and I use it because I find it easier to digest when the spec doesn't run immediately into the code

JeetKunDoug: Problem is it's _not_ common in Basho code - we pretty much universally do it the other way (comments then spec then function) - consistency with the rest of the codebase is important. Same goes for 2 `%` symbols for left-aligned comments (even when not `@doc` comments) - I've almost never seen single-`%` comments in our codebase, which is why it stuck out at me like a sore thumb.

tburghart: I refer you to the edoc [documentation](http://erlang.org/doc/apps/edoc/chapter.html#Type_specifications):

> Examples:
```
-spec my_function(X :: integer()) -> integer().
%% @doc Creates ...
```
> The first example shows the recommended way of specifying functions.

There may be more of it in our source base than you think - it's not all from me.

tburghart: The latest commit takes care of the comment format.

JeetKunDoug: Please go take a look at almost any other use of a `-spec` and a comment above a function in our codebase (specifically, modules starting with `riak_` since we "own" those more than others)... In every case I've looked at (and I'm at about 100 instances of `-spec` so far) it's doc/spec/function. I think keeping consistent with the rest of our codebase is important enough to spend some time swapping these around.

JeetKunDoug:
+1

Issue #238, opened 2017-01-28
Build failed on Mac OS 10.11.6  [JIRA: RIAK-3222]


fogfish:
The latest commit causes build failure on Mac OS 10.11.6
 
https://github.com/basho/eleveldb/commit/08ea4821a174aa6fbfd0d1b54520687716d323d5

```
ld: warning: object file (/Users/dkolesnikov/devel/z/typhoon/_build/default/lib/eleveldb/c_src/system/lib/libsnappy.a(snappy.o)) was built for newer OSX version (10.11) than being linked (10.8)
ld: warning: object file (/Users/dkolesnikov/devel/z/typhoon/_build/default/lib/eleveldb/c_src/system/lib/libsnappy.a(snappy-sinksource.o)) was built for newer OSX version (10.11) than being linked (10.8)
Undefined symbols for architecture x86_64:
  "std::__1::locale::use_facet(std::__1::locale::id&) const", referenced from:
      std::__1::basic_ostream<char, std::__1::char_traits<char> >& std::__1::__put_character_sequence<char, std::__1::char_traits<char> >(std::__1::basic_ostream<char, std::__1::char_traits<char> >&, char const*, unsigned long) in libsnappy.a(snappy.o)
      snappy::LogMessageCrash::~LogMessageCrash() in libsnappy.a(snappy.o)
      snappy::LogMessage::~LogMessage() in libsnappy.a(snappy.o)
  "std::__1::ios_base::getloc() const", referenced from:
      std::__1::basic_ostream<char, std::__1::char_traits<char> >& std::__1::__put_character_sequence<char, std::__1::char_traits<char> >(std::__1::basic_ostream<char, std::__1::char_traits<char> >&, char const*, unsigned long) in libsnappy.a(snappy.o)
      snappy::LogMessageCrash::~LogMessageCrash() in libsnappy.a(snappy.o)
      snappy::LogMessage::~LogMessage() in libsnappy.a(snappy.o)
  "std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >::__init(char const*, unsigned long)", referenced from:
      snappy::internal::WorkingMemory::GetHashTable(unsigned long, int*) in libsnappy.a(snappy.o)
      snappy::internal::CompressFragment(char const*, unsigned long, char*, unsigned short*, int) in libsnappy.a(snappy.o)
      snappy::Compress(snappy::Source*, snappy::Sink*) in libsnappy.a(snappy.o)
[...]

PR #241, opened 2017-02-02
Added flag for MACOSX_DEPLOYMENT_TARGET = 10.8

angrycub:
This value is an alternative to `-mmacosx-version-min=version` which
does not appear to be settable via the snappy configure script. This is
to allow support for historical versions of MacOSX


angrycub:
This fixes https://github.com/basho/eleveldb/issues/238 and has been verified by the reporter.

PR #243, opened 2017-12-07
Add a 10% fuzz factor to the resident memory calc (intermittent test failure "fixes")

russelldb:
On some platforms (or on ubuntu-16.04 server in my lab at least) this
test fails going a little over the expected max resident
memory. However, even with loads more loops, a 10% fuzz factor is enough
for it to pass. If there is a leak, it is a marginal one that never
grows above 10% of the best case, which is good enough, since this code
has been in the wild long enough.


martincox:
+1.

PR #244, opened 2018-02-28
Feature/use system snappy - don't merge - needs more work

ghost:
Use system snappy lib

We were shipping an old version of snappy, storing it as a tar file, extracting, building the lib and then staticly compiling it into our eleveldb.so.

Why do this? Snappy is a library (snappy) via brew on OSX, via apt-get/yum (libsnappy-dev).

Our version is also old, and reduces farm to plate traceability.

This changeset removes the bundled tar file (c_src/snappy-1.0.4.tar.gz) and uses the system lib instead.


Licenser:
👍 works on SmartOS

ghost:
Superceeded by #254

PR #248, opened 2019-03-14
Replicate build changes

martinsumner:
Compatible with MAC OSX as well as always taking the leveldb_ee version


martinsumner:
Currently 2.9.0 points at github/martinsumner - want to make the changes necessary to point back at github/basho.

This requires the build fixes for OSX and always leveldb_ee to be present on a develop-2.9 branch

PR #254, opened 2019-08-14
 rebar3 compile on osx/debian/alpine

ghost:
Lets operate on the basis that : 

1. The future of Riak is leveled - particularly as there's been next to no maintenance on this since 2016 
2. The develop branch is for users of rebar3, that's a reasonable presumption in 2019
3. People would like to be able to build on OSX, Alpine, Debian.
4. People are using `riak_core` and it needs hashtree/metadata etc which are all dependent upon eleveldb (for now).
5. People are using Erlang 20+, and in many cases Erlang 22 
6. People are using `riak_core` in Elixir applications
7. Lots of people do their daily work on OSX.

This PR : 

1. Removes the usage of rebar (the binary blob that you don't really know where it came from or what it does).
2. Provides two dockerfiles, one for Alpine, one for debian that demonstrate what's necessary to get eleveldb to build on those two platforms, using rebar3, with these changes - perhaps those docker builds could be added to travis (if it's still running) in a future PR.
3. Uses the operating system provided version of snappy instead of using an arbitrary, unverified (no checksum) version of the snappy library marked as 1.0.4 - incidentally, there is an existing security advisory for snappy version 1.1.4. https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2018-7577 - in any event the intention was to switch to lz4 as the default - I think we can agree this doesn't represent very good practice now (made sense at the time).

Downside : 
The tools for inspecting SST files don't build - can't get them to work in any of the envs (OSX,Alpine,Debian) - perhaps someone with better c/c++ experience can figure that out - in any event nobody uses them apart from @matthewvon - maybe he can offer a suggestion/tweak to get them working again. I think that's acceptable for now.


martinsumner:
On other Basho repos @nsaadouni has done a tidy-up of branches and made develop-3.0 (i.e. the OTP 20 target branch) the default.  Would this be helpful here too?  With the docker support then merged into develop-3.0?

robertoaloi:
Hi @bryanhuntesl do you know what is the plan with this PR? I'm mostly interested in being able to compile `eleveldb` with rebar3 on OS X.

ghost:
It joins the rest of the list of lose ends on the risk project. maybe fork using this branch ? @robertoaloi

PR #255, opened 2019-10-04
Mas 3.0 dialyzer1

martinsumner:
More dialyzer fixes

### DIFF
--- a/src/eleveldb.erl
+++ b/src/eleveldb.erl
@@ -139,7 +139,7 @@ init() ->
 -type read_options() :: [read_option()].
 
 -type fold_option()  :: {first_key, Key::binary()}.
--type fold_options() :: [read_option() | fold_option()].
+-type fold_options() :: [fold_option() | read_options()].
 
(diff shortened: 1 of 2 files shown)


martinsumner:
@ThomasArts

src/eleveldb.erl
```diff
@@ -132,14 +132,14 @@ init() ->
                          {whole_file_expiry, boolean()}
                         ].
 
--type read_option() :: {verify_checksums, boolean()} |
-                       {fill_cache, boolean()} |
-                       {iterator_refresh, boolean()}.
+-type read_options() :: [{verify_checksums, boolean()} |
```
ThomasArts, line 135: This is equivalent... but I preferred the original notation.
But don't worry about it.

[ a | b ] can easily be read as a list with either an a or a b.
whereas list [ a_or_b() ] is a list of a's and b's (plural) and even the empty list is a list that fits.
But, semantically this is equivalent.

ThomasArts (approved):

PR #259, opened 2020-02-17
Bump leveldb version

martinsumner:
To one which builds on 32-bit platforms


Bob-The-Marauder:
+1 - been waiting for this a long time

PR #261, opened 2020-05-07
Develop 3.0 292

martinsumner:
Standardise versions and tidy-up


c_src/build_deps.sh
```diff
@@ -8,7 +8,7 @@ if [ `uname -s` = 'SunOS' -a "${POSIX_SHELL}" != "true" ]; then
 fi
 unset POSIX_SHELL # clear it so if we invoke other scripts, they run as ksh as well
 
-LEVELDB_VSN="2.0.35"
+LEVELDB_VSN="2.0.36"
```
martincox, line 11: What was the reason for bumping this up? We've had some odd leveldb stuff occur, so are quite weary of taking any changes, especially considering the code is essentially an basically an untaintained black box.

martincox (approved):

c_src/build_deps.sh
martinsumner: Just the 32-bit build support (as discussed).  Will go forward with this, and leave out the ARM64 PR for now

PR #263, opened 2022-01-11
Updated the url schema due to new Github security policy

phiotr:
Updated the url schema due to new Github security policy described below:
https://github.blog/2021-09-01-improving-git-protocol-security-github/


phiotr:
The same Pull Request from another user was approved. So closing my copy as a duplicate.

PR #264, opened 2022-01-11
Fix git clone URL

om26er:
Starting today, we can no longer clone a repository from Github using `git://` protocol.
```
om26er@HomePC:~/scm/simplethings/eleveldb$ git clone git://github.com/basho/leveldb
Cloning into 'leveldb'...
fatal: remote error: 
  The unauthenticated git protocol on port 9418 is no longer supported.
Please see https://github.blog/2021-09-01-improving-git-protocol-security-github/ for more information.
```

This PR changes the code to use https instead.

### DIFF
--- a/c_src/build_deps.sh
+++ b/c_src/build_deps.sh
@@ -56,7 +56,7 @@ case "$1" in
 
     get-deps)
         if [ ! -d leveldb ]; then
-            git clone git://github.com/basho/leveldb
+            git clone https://github.com/basho/leveldb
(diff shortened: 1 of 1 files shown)


martinsumner (approved):

martinsumner:
Only `develop-3.0` branch is being actively developed (and this, I believe has this change already) - and even then only small, critical fixes.  It might be better to switch to `develop-3.0`, rather than trying to update `develop`.

martincox:
Yep, already fixed in develop-3.0.

PR #271, opened 2023-02-16
Depend on system snappy

hmmr:
Rather than downloading and building snappy, depend on libsnappy preinstalled on the host system.

Also, use `--depth=1` when cloning leveldb.

### DIFF
--- a/.github/workflows/erlang.yml
+++ b/.github/workflows/erlang.yml
@@ -29,7 +29,7 @@ jobs:
         run: apt-get update && apt-get install -y cmake && make
       - name: Run tests
         run: apt-get update && apt-get install -y cmake libc6-dev && make check
-      - name: rebar3 compile 
+      - name: rebar3 compile
         run: ./rebar3 compile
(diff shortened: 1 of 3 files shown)


codeadict:
I'm getting ` ld: library not found for -lsnappy` on an ARM mac by installing snappy from homebrew. I'll do more investigation this weekend but seems like the problem doesn't get fixed for OSX

ioolkos:
@codeadict I think we had a couple of commits over at the VerneMQ fork, related to building on FreeBSD and Darwin. Not sure this is directly related, but maybe there's some stuff useful to upstream or cherrypick. https://github.com/vernemq/eleveldb/pull/10

hmmr:
For the record, on FreeBSD-13.1, after `pkg install snappy`, it is required to `export LDFLAGS=-L/usr/local/lib` before running `make`.

hmmr:
Closing, in favor of #272.

hmmr:
Closing, in favor of #272.

PR #274, opened 2023-05-02
Introduce rebar3_external plugin to handle leveldb source code

joaohf:
Currently [1], rebar3 does not support adding hooks for 'get-deps'. That is why the compile rule in c_src/Makefile has also to call get_deps makefile rule in order to fetch leveldb source code. The whole process works but can be improved.

This change introduces a plugin called rebar3_external which deals with external source code like leveldb source code. That way rebar3 can perform dependency manager without relies on makefiles or scripts.

The intend is to have leveldb source code managed by rebar3. So, one could easily upgrade it just bumping the leveldb version in the rebar3.config file.

1: https://github.com/erlang/rebar3/issues/2784

### DIFF
--- a/c_src/Makefile
+++ b/c_src/Makefile
@@ -1,4 +1,4 @@
-LEVELDB_VSN ?= "2.0.38"
+LEVELDB_SRC := $(REBAR_DEPS_DIR)/leveldb/c_src
 C_SRC_DIR := $(shell pwd)
 PROJECT_DIR := $(abspath $(C_SRC_DIR)/..)
 TARGET_DIR := $(PROJECT_DIR)/priv/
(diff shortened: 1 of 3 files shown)


joaohf:
Hi,

I'm sharing this work as a try to improve some parts of riak build. Mainly to address some points detailed here: https://github.com/erlang/rebar3/issues/2784

tburghart:
@joaohf There is a fully-tested, though out of date, implementation of this strategy at https://github.com/basho/eleveldb/tree/294dc67687590d6c830c09cbefbbec912c076069 (was the head of the `rebar3` branch - the branch has been deleted, but the commits are still there).

The `basho_rebar_tools` plugin mentioned in `rebar.config` is not needed for the build and can simple be deleted. Substituting `alertlogic`'s fork of the `rebar_raw_resource` plugin _should_ make the whole thing work.

This was a complete rewrite of the `eleveldb` build to fix a multitude of sins, not least of which was treating `leveldb` and `snappy` as proper rebar-managed dependencies (`leveldb_ee` should be added to that list now that it's open source). At the time, it supported building on Basho's OTP R16 version and everything later, though of course rebar3 itself no longer supports that.

joaohf:
Hi @tburghart 

Now I see. I didn't check all branches from eleveldb before. Also the rebar_raw_resource also fixes my initial point (which is more related to rebar3 dependency life cycle, for example fetching all deps when calling get-deps).

So, just asking for my own understandment, why the rebar3 branch has not been merged at that time ?

Thanks.

tburghart:
> why the rebar3 branch has not been merged at that time ?

I can't answer that, I wasn't involved in that decision. Had Basho not folded, it would have gone into the OTP 20+ version of Riak, but things kinda went off the rails ...

Issue #275, opened 2023-07-07
Snappy support is broken


benjoe87:
Snappy support is broken since https://github.com/basho/eleveldb/commit/6ef920211272a288316f0ffa21cb79350e63effc.

The root cause seems like:
1. The build flags are not exported in the Makefile 
2. The leveldb build cannot detect the snappy support and don't set the SNAPPY C marco https://github.com/basho/leveldb/blob/develop/build_detect_platform#L169
3. The leveldb part is builded without snappy support.

It is hard to detect this issue during the tests because leveldb simply do not compress the blocks if snappy support is not present.
https://github.com/basho/leveldb/blob/b91d5cee860bc708c10c48911586fdb347558ea3/table/table_builder.cc#L212

But it definitely breaks the upgrades mentioned in https://github.com/basho/eleveldb/issues/273

Exporting these variables probably solves the issue but I am not 100% sure that this is a comprehensive solution.

Issue #273, opened 2023-04-12
vnode hangs in `{eleveldb,write,3}`


systream:
In RIAK 3.0.12, vnodes started to hang in leveldb:write/3 and the process mailbox could grow up to 10K messages.
After a while whole node starts to behave unstable. The queue size of the vnode does not decrease, even without a load.

```
[{current_function,{eleveldb,write,3}},
 {initial_call,{proc_lib,init_p,5}},
 {status,waiting},
 {message_queue_len,10001},
 {links,[<0.1687.0>,<0.2146.0>,<0.258.0>]},
 {dictionary,[{rand_seed,{#{bits => 58,jump => #Fun<rand.13.8986388>,
                            next => #Fun<rand.10.8986388>,type => exsss,
                            uniform => #Fun<rand.11.8986388>,
                            uniform_n => #Fun<rand.12.8986388>},
                          [112506433278832578|159401657947672872]}},
              {#Ref<0.2265291839.1457520643.98734>,
               {bc_state,"/var/lib/riak/bitcask/11417981541647679048466287755595961091061972992",
                         fresh,undefined,[],2147483648,
                         [{expiry_secs,-1},
                          {io_mode,erlang},
                          {expiry_grace_time,0},
                          {small_file_threshold,10485760},
                          {dead_bytes_threshold,134217728},
                          {frag_threshold,40},
                          {dead_bytes_merge_trigger,536870912},
                          {frag_merge_trigger,60},
                          {max_file_size,2147483648},
                          {open_timeout,4},
                          {data_root,"/var/lib/riak/bitcask"},
                          {sync_strategy,none},
                          {merge_window,always},
                          {max_fold_age,-1},
                          {max_fold_puts,0},
                          {expiry_secs,-1},
                          {require_hint_crc,true},
                          {key_transform,#Fun<riak_kv_bitcask_backend.1.43379991>},
                          {read_write,true}], 
                         #Fun<riak_kv_bitcask_backend.1.43379991>,
                         #Ref<0.2265291839.1457651715.98733>,1,2}},
              {'$initial_call',{riak_core_vnode,init,1}},
              {'$ancestors',[riak_core_vnode_sup,riak_core_sup,<0.254.0>]},
              {bitcask_time_fudge,no_testing},
              {hashtree_tokens,47},
              {#Ref<0.2265291839.1457520644.95496>,
               {bc_state,"/var/lib/riak/bitcask/11417981541647679048466287755595961091061972992",
[...]

PR #276, opened 2023-07-07
Fix broken snappy support in leveldb

benjoe87:
Without exporting the compiling flags the build_detect_platform is not able to detect the snappy support and does not set the SNAPPY C macro.

Fix issue https://github.com/basho/eleveldb/issues/275

### DIFF
--- a/c_src/Makefile
+++ b/c_src/Makefile
@@ -1,11 +1,11 @@
-LEVELDB_VSN ?= "2.0.38"
+export LEVELDB_VSN ?= "2.0.38"
 SNAPPY_VSN ?= "1.1.9"
 BASEDIR = $(shell pwd)
 
-LDFLAGS := $(LDFLAGS) -L$(BASEDIR)/system/lib
-LD_LIBRARY_PATH := $(BASEDIR)/system/lib:$(LD_LIBRARY_PATH)
-CFLAGS := $(CFLAGS) -I $(BASEDIR)/system/include -I. -I $(BASEDIR)/leveldb/include -fPIC
-CXXFLAGS := $(CXXFLAGS) -I $(BASEDIR)/system/include -I. -I $(BASEDIR)/leveldb/include -fPIC
+export LDFLAGS := $(LDFLAGS) -L$(BASEDIR)/system/lib
+export LD_LIBRARY_PATH := $(BASEDIR)/system/lib:$(LD_LIBRARY_PATH)
+export CFLAGS := $(CFLAGS) -I $(BASEDIR)/system/include -I. -I $(BASEDIR)/leveldb/include -fPIC
+export CXXFLAGS := $(CXXFLAGS) -I $(BASEDIR)/system/include -I. -I $(BASEDIR)/leveldb/include -fPIC
 
 get-deps:
 	if [ ! -r snappy-$(SNAPPY_VSN).tar.gz ]; then \


systream:
@martinsumner how could we proceed with this. 
We still using leveldb. :/

martinsumner:
I wanted to have a test to make sure if we do fix this issue, we're not going to regress the fix.

This [updated riak_test upgrade test](https://github.com/basho/riak_test/pull/1382) detects both the problem of compression not working, as well as the [previously undetected issue](https://github.com/basho/eleveldb/issues/273) of snappy updates being uncompatable.

However, when I test this on a non-update (e.g. previous and current both `develop-3.0`) but with the eleveldb branch switched to  `benjoe87:fix-snappy-support` - the test still failed.  The size of the backend store didn't reduce when comparing snappy with compression disabled:

```
================ verify_basic_upgrade failure stack trace =====================
{{assert,[{module,verify_basic_upgrade},
          {line,115},
          {expression,"NativeSz < floor ( ( PlainTextSz * 0.8 ) )"},
          {expected,true},
          {value,false}]},
 [{verify_basic_upgrade,'-confirm/0-fun-12-',2,
                        [{file,"/Users/martinsumner/riak_test/tests/verify_basic_upgrade.erl"},
                         {line,115}]},
  {verify_basic_upgrade,confirm,0,
                        [{file,"/Users/martinsumner/riak_test/tests/verify_basic_upgrade.erl"},
                         {line,115}]},
  {riak_test_runner,return_to_exit,3,
                    [{file,"/Users/martinsumner/riak_test/src/riak_test_runner.erl"},
                     {line,159}]}]}
===============================================================================
```

So I can't seem to get this PR to resolve the problem.  Is there other evidence this works?

martinsumner:
@tburghart is there a fix for this in your version?

martinsumner:
Previously when I tested this fix, using the [updated verify_basic_upgrade](https://github.com/basho/riak_test/pull/1382) - it FAILED.  But this was testing on my OSX laptop.

I tried this fix today on Ubuntu - and it worked.

- setting `current` 3.0.16, `previous` 3.0.16 -> `verify_basic_upgrade` **FAIL** (as size of leveldb store not reduced by enabling compression).
- setting `current` 3.0.16 + ELEVELDB PR276, `previous` 3.0.16 + ELEVELDB PR276 -> `verify_basic_upgrade` **PASS** (compression now enabled)
- setting `current` 3.0.16 + ELEVELDB PR276, `previous` 3.0.11 -> `verify_basic_upgrade` **PASS** (1.1.9 and 1.04 are in fact backwards/forwards compatible)

This would appear to confirm this PR is good on Ubuntu.  The original [issue](https://github.com/basho/eleveldb/issues/273) is caused by snappy not running and not due to incompatibility between 1.0.4 and 1.1.9.

As OSX is not supported for running production Riak clusters, I think the PR should be accepted so that there can be a 3.0.17 release that will allow multi_backend users (or those who have over-ridden the LZ4 default) to safely upgrade from 3.0.11.

@nsaadouni - is this something you can look at?

martinsumner:
The OSX issues can be resolved by setting "std=c++11" in the CFLAGS and COMMON_FLAGS in build_detect_platform:

https://github.com/nhs-riak/leveldb/commit/458bfb4e41d2822b9f2e90b393ec786689249c33

Also in the rebar.config.script so that it is used when compiling eleveldb:

https://github.com/nhs-riak/eleveldb/commit/50dd137fd8aa00a4af71d3c11af9528b0a2f40c3#diff-a07a3d3b357a34c4acb80b2e178e9a19e6c2ed9d91cb0aa8d53cadabb1ae8bd0

tburghart:
Sorry for being late to the discussion.
The way we've "fixed" it in our version is to completely rewrite `eleveldb/c_src/Makefile` to:
* clone the leveldb, leveldb_ee, and snappy repos directly by version (no submodules)
* run (and clean up after) snappy's cmake
* build the leveldb archive and tools using a tuned-up version of `leveldb/build_detect_platform`

This has given us not only a clean build across platforms, but also a tree that recognizes it's already been built so it doesn't rebuild 8 times when making devrels 🤨

